PR #18, opened 2020-11-12
Route accessibility labels in Compose

justasm:
Good resources on semantics in Compose:
- https://bryanherbst.com/2020/10/12/compose-semantics-intro/
- https://bryanherbst.com/2020/11/03/compose-semantics-talkback/

### DIFF
--- a/android/routes/ui/src/main/java/com/trafi/routes/ui/Route.kt
+++ b/android/routes/ui/src/main/java/com/trafi/routes/ui/Route.kt
@@ -13,16 +13,29 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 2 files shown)


android/routes/ui/src/main/java/com/trafi/routes/ui/Route.kt
```diff
@@ -13,16 +13,29 @@ import androidx.compose.material.Text
...
+import java.text.DateFormat
+import java.util.Date
+import java.util.TimeZone
 
 @Composable
 fun Route(route: Route, modifier: Modifier = Modifier) {
-    Surface(modifier = modifier.clickable(onClick = {})) {
+    Surface(modifier = modifier
+        .semantics(mergeAllDescendants = true) {
+            accessibilityLabel = route.accessibilityLabel
```
justasm, line 36: Equivalent to `android:contentDescription`

PR #21, opened 2020-11-15
Android build action

absimas:
Create a github action to check android modules and assemble the demo app.
Skipping ktlint from `common` library since model generation is not complete there yet.

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -1,0 +1,46 @@
+name: Build Android
+on:
+  workflow_dispatch:
+  push:
+    branches: master
+  pull_request:
+    branches: master
+jobs:
+  Build:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 12
+      - name: ktlint
+        if: always()
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: ktlintCheck
+      - name: Lint
+        if: always()
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: lint
+      - name: Test
+        if: always()
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: test
+      - name: Assemble
+        if: always()
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: assemble
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -30,6 +30,11 @@ jobs:
           build-root-directory: android
           wrapper-directory: android
           arguments: lint
+      - name: Retain lint artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: lint.html
+          path: android/routes/ui/build/reports/lint-results.html
       - name: Test
         if: always()
         uses: eskatos/gradle-command-action@v1
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Retain lint artifacts
         uses: actions/upload-artifact@v2
         with:
-          name: lint.html
+          name: lint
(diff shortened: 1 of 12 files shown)


.github/workflows/build_android.yml
```diff
@@ -0,0 +1,46 @@
+name: Build Android
+on:
+  workflow_dispatch:
+  push:
+    branches: master
```
justasm, line 5: The primary branch is `main`

.github/workflows/build_android.yml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:
+    branches: master
```
justasm, line 7: 🔝

justasm (commented):
Awesome! 👏

I'm trying to understand whether the build environment will have the necessary Android build tools 🤔 Have you found any good resources for good practices with GitHub Actions?

README.md
```diff
@@ -1,4 +1,6 @@
-## MaaS Components
+# MaaS Components
+
+[![Android Build](https://github.com/trafi/maas-components/workflows/Android%20Build/badge.svg?event=push)](https://github.com/trafi/maas-components/actions?query=workflow%3A%22Android+Build%22)
```
justasm, line 3: This badge doesn't currently work, will it? 🙂

absimas: Should work after merge. I'm 90% sure it's correct :smile:

absimas: Well that 10% got me :smile: Action appeared before merging once started using `main` branch :pray:

absimas:
Re: good resources.

Their [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions) seems to be quite informative though a bit extensive.
Also a good reference is a list of their [virtual environments](https://github.com/actions/virtual-environments#available-environments). Which also have the build tools listed.

.github/workflows/build_android.yml
```diff
@@ -0,0 +1,46 @@
...
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: ktlintCheck
+      - name: Lint
+        if: always()
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: lint
```
justasm, line 32: How can we see the Lint artifacts in the workflow results? Currently it doesn't seem to list them

absimas: They need to be specifically retained. Not sure if that's necessary since we see the logs anyway.

.github/workflows/build_android.yml
```diff
@@ -0,0 +1,46 @@
...
+        os: [ ubuntu-latest, macos-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 12
+      - name: ktlint
+        if: always()
+        uses: eskatos/gradle-command-action@v1
```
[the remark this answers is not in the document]
absimas: Mostly google + most starred action.

.github/workflows/build_android.yml
```diff
@@ -0,0 +1,46 @@
...
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: test
+      - name: Assemble
+        if: always()
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: android
+          wrapper-directory: android
+          arguments: assemble
```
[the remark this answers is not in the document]
absimas: It depends on how long GitHub actions are free. For private use MacOS agents are 10 times more expensive than Linux agents. In such a case I would dedicate as little work to MacOS agents as possible.

At the moment, however, we have the luxury of making sure our builds are still compatible on both platforms.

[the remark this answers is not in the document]
justasm: My confusion is that we're building for the iOS targets in an action called "Build Android" 🙂

.github/workflows/build_android.yml
justasm: If it doesn't cost us anything, it'd be great to retain the Lint reports! The HTML reports are much easier to browse than spelunking the build logs.

justasm (approved):

.github/workflows/build_android.yml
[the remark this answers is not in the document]
absimas: Got it. It wasn't clear to me how KMP will be built in the future.
Will it have a separate action? Will it be bundled differently based on the executed platform?

Because of these questions (and the desire to keep these changes small) I decided it would be fine to bundle KMP builds into Build Android's action.

In my opinion, once we have a clearer distinction on what's built when we can expand/split this up a bit more.

PR #22, opened 2020-11-18
Add Maven Central badge

justasm:
Using [shields.io](http://shields.io).

Unfortunately, we don't currently have an umbrella artifact for all dependencies, i.e. `locations-android`, `routes-ui-android`, etc. Using the `routes-ui-android` artifact for versioning for now.


absimas (commented):
Cool!
Side note: I was wondering yesterday.. Should the android badge be in the Android's README? :thinking:

justasm:
Yeah, I was wondering too 🙂 I think either or works. Why not both? 😄

PR #26, opened 2020-11-27
Bootstrap iOS snapshot testing

domasn:
It's pretty neat! We came up with this during QA day - see [summary in Slack](https://trafi.slack.com/archives/CJR57M6GY/p1605714596015600).

We're using [swift-snapshot-testing](https://github.com/pointfreeco/swift-snapshot-testing) by pointfreeco 🙌 

We can even see the diff right in GitHub 😮 
You can see 👇 by looking at rich diff of this https://github.com/trafi/maas-components/pull/26/commits/45cb6cfcd668364666371ac8ec82216e852af7d7#diff-1edd10da282cfec29ef4950912986fcad8c503444f2fd84b3058dbab0b17d013
image

⚠️ `swift test` does not work as it tries to run on macOS architecture instead of iOS - sorting it out 😬 [This looks](https://arjunnayini.com/2019-11-01-github-ci-swift-package/) a bit much, but promissing.

cc @aistetests @eimantas  @justasm


aistetests:
Tests fail for me :( It seems the resolution or compression of the image is different. Here's how it looks:

[Screen Recording 2020-11-27 at 17.05.33.zip](https://github.com/trafi/maas-components/files/5608445/Screen.Recording.2020-11-27.at.17.05.33.zip)

domasn:
@aistetests you're correct! We should all be running our tests on @2x iPhones 😬 
https://github.com/pointfreeco/swift-snapshot-testing/issues/165#issuecomment-455173203

PR #29, opened 2020-12-07
Analytics in common:core module

justasm:
Those with access to Trafi's internal VPN can [access & edit the event schema via the Mammoth editor](https://mammoth.trafi.com/open-sdk/events).

Uses https://github.com/trafi/mammoth-kt to generate type-safe analytics events from the schema.

Any UI component can call `Analytics.consume(AnalyticsEvent.$eventName())`. By default, events go nowhere. Any SDK consumer can register an `AnalyticsDelegate` to do something with the events.

```kotlin
class PrintAnalyticsToConsole : AnalyticsDelegate {
  fun dispatchEvent(event: Analytics.Event) {
    println(event.business.name)
  }

  val screenName: String? get() = null
  val modalName: String? get() = null
}

Analytics.set(PrintAnalyticsToConsole())
```

### DIFF
--- a/common/core/src/commonMain/kotlin/com/trafi/analytics/Analytics.kt
+++ b/common/core/src/commonMain/kotlin/com/trafi/analytics/Analytics.kt
@@ -1,0 +1,31 @@
+package com.trafi.analytics
+
+import kotlin.native.concurrent.ThreadLocal
+
+interface AnalyticsDelegate {
+    fun dispatchEvent(event: Analytics.Event)
+
+    val screenName: String?
+    val modalName: String?
+}
+
+@ThreadLocal
+object Analytics {
+
+    private var delegate: AnalyticsDelegate? = null
+
+    val screenName: String get() = delegate?.screenName ?: ""
+    val modalName: String get() = delegate?.modalName ?: ""
+
+    fun consume(event: Event) {
+        delegate?.dispatchEvent(event)
+    }
+
+    fun set(delegate: AnalyticsDelegate) {
+        this.delegate = delegate
+    }
+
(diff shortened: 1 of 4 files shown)


common/core/src/commonMain/kotlin/com/trafi/analytics/Analytics.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    fun consume(event: Event) {
+        delegate?.dispatchEvent(event)
+    }
+
+    fun set(delegate: AnalyticsDelegate) {
+        this.delegate = delegate
+    }
+
+    data class Event(val business: RawEvent?, val publish: RawEvent)
```
justasm, line 28: `business` should not be nullable

common/core/src/commonMain/kotlin/com/trafi/analytics/Analytics.kt
```diff
@@ -0,0 +1,31 @@
...
+
+interface AnalyticsDelegate {
+    fun dispatchEvent(event: Analytics.Event)
+
+    val screenName: String?
+    val modalName: String?
+}
+
+@ThreadLocal
+object Analytics {
```
eimantas, line 13: It would be great to have all these properties and functions static. In Swift we consume events through a static method: `Analytics.consume(event)`. Currently the generated interface requires to instantiate the `Analytics` type and call methods from there.

justasm: Please experiment to see what Kotlin code gives you the result you expect. Ping me if you'd like to brainstorm ideas together.

PR #32, opened 2020-12-07
Support themed colors

eimantas:
Usage involves:

```
let defaultColor = Color.primary // uses current environment tint
let forcedLightColor = Color.primary(.light)
let forcedDarkColor = Color.primary(.dark)
```

### DIFF
--- a/ios/Sources/MaaSTheme/Colors.swift
+++ b/ios/Sources/MaaSTheme/Colors.swift
@@ -2,7 +2,7 @@ import SwiftUI
 import UIKit
 import MaasCore
 
-extension Color {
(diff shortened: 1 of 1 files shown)


justasm:
Can't comment much as this is specific to iOS & SwiftUI API conventions, but looks reasonable 👍

On Android & Compose we don't have the concept of light / dark for individual colors - instead, the API user specifies the entire color palette and can specify a different one in the case of a dark theme. This is consistent with Google's Material theme implementation in Compose.

justasm:
Seems to be implemented in #37 and #39? @eimantas

PR #34, opened 2020-12-07
Button constants in Kotlin/Native

justasm:
```kotlin
class ButtonConstants(theme: CurrentTheme) {
     val defaultColor = theme.colorPalette.primary
     val defaultContentColor = theme.colorPalette.onPrimary
     val disabledColor = Gray200
     val textStyle = theme.typographyScale.textL.copy(fontWeight = Bold)
     val minHeight = 48.dp
     val cornerRadius = theme.cornerRadiusScale.buttonRadius
     val defaultTruncate = true
 }
```

Depends on https://github.com/trafi/maas-components/pull/37

cc @domasn


common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/ButtonConstants.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.trafi.ui.theme.internal.BasicFontWeight.*
+import com.trafi.ui.theme.internal.ColorPalette.Gray200
+import com.trafi.ui.theme.internal.CurrentTheme
+import com.trafi.ui.theme.internal.copy
+
+class ButtonConstants(theme: CurrentTheme) {
+    val defaultColor = theme.colorPalette.primary
+    val defaultContentColor = theme.colorPalette.onPrimary
+    val disabledColor = Gray200
```
eligijusTrafi, line 11: Should't this color be different for dark theme?

eligijusTrafi: And it is not clear if this is disabled button background color or  disabled button content color

justasm: Good questions 🙂 Just plain "color" refers to the background color, although we could definitely be more explicit if we want to.

eligijusTrafi: No, its ok to keep this naming convention 👍 Personally to me, disabled content color would be nice as white content color on a light gray background color doesn't seem to be nice(Specially when i'm not wearing my 👓) 😄 
Screen Shot 2020-12-10 at 16 34 59

mikasdo (approved):

eligijusTrafi (approved):

PR #35, opened 2020-12-08
Support for logging network requests

TomasSkar:
Add support for logging all network requests.

If a `Logger` is specified, all network requests & responses are logged.

```kotlin
data class ApiConfiguration(
    // ..
    val logger: Logger? = null,
)
```

Alternatives considered and rejected:
- [Napier](https://github.com/AAkira/Napier) internally uses mutable state and requires setup during initialization on both platforms, does not properly manage Kotlin/Native concurrency.
- [Kermit](https://github.com/touchlab/Kermit) requires transitive export of dependencies and does not seem to be very actively maintained.
- Logging logic within the shared Kotlin multiplatform logic cannot yet use `os_log` on Apple platforms.

image


common/core/src/commonMain/kotlin/com/trafi/core/HttpClient.kt
```diff
@@ -7,9 +7,14 @@ import io.ktor.client.features.json.serializer.KotlinxSerializer
 import io.ktor.client.request.header
 import io.ktor.http.ContentType
 import io.ktor.http.contentType
+import io.ktor.client.features.logging.*
 import kotlinx.serialization.json.Json
 
 internal fun httpClient(apiKey: String) = HttpClient {
+    install(Logging) {
+        logger = Logger.DEFAULT
+        level = LogLevel.ALL
```
justasm, line 16: We need a way to avoid enabling logging of everything for all SDK users.

justasm (changes requested):
Really happy to see this initiative, this is one thing we need to solve 👍 We need to make this configurable in some way such that
1. the SDK does not log anything (except for incorrect API usage) by default
1. logging can be enabled / disabled by the SDK user
1. ~logging can be customized by the SDK user (nice-to-have)

common/core/src/commonMain/kotlin/com/trafi/core/ApiConfiguration.kt
```diff
@@ -1,7 +1,12 @@
...
 
+/**
+ * [getIdToken] - lambda that should always return the latest id token. Used for authenticated requests.
+ * [logger] - if specified, logs all network requests & responses.
+ */
 data class ApiConfiguration(
     val baseUrl: String,
     val apiKey: String,
     val getIdToken: () -> String? = { null },
+    val logger: Logger? = null,
```
justasm, line 11: This is the key API change.

justasm (commented):
Tested with the Android sample app, working as expected 👍

justasm (approved):

Ignelios (approved):
LGTM

PR #37, opened 2020-12-10
Current theme in common:core module

justasm:
Enable Kotlin/Native code to be aware of and use values from the current theme. Enables theme-aware code to be shared across iOS & Android via Kotlin/Native; see #34 for an example. To enable bidirectional mapping, text styles in common:core are now declared using expect / actual and map directly to Compose `TextStyle` / iOS `UIFont`.

```kotlin
expect class OsTextStyle

// android
actual typealias OsTextStyle = TextStyle

// ios
actual typealias OsTextStyle = UIFont
```

cc @domasn

### DIFF
--- a/android/ui/src/main/java/com/trafi/ui/theme/CurrentTheme.kt
+++ b/android/ui/src/main/java/com/trafi/ui/theme/CurrentTheme.kt
@@ -1,0 +1,63 @@
+package com.trafi.ui.theme
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.unit.Dp
+import com.trafi.ui.theme.internal.CurrentColorPalette
+import com.trafi.ui.theme.internal.CurrentCornerRadiusScale
+import com.trafi.ui.theme.internal.CurrentSpacingScale
+import com.trafi.ui.theme.internal.CurrentTheme
+import com.trafi.ui.theme.internal.CurrentTypographyScale
+
+@Composable
+internal val currentTheme: CurrentTheme
+    get() {
+        val colors = MaasTheme.colors
+        val typography = MaasTheme.typography
+        val spacing = MaasTheme.spacing
+        val cornerRadius = MaasTheme.cornerRadius
+        return CurrentTheme(
+            colorPalette = with(colors) {
+                CurrentColorPalette(
+                    primary = primary.osColor,
+                    primaryVariant = primaryVariant.osColor,
+                    secondary = secondary.osColor,
+                    secondaryVariant = secondaryVariant.osColor,
+                    background = background.osColor,
+                    surface = surface.osColor,
+                    error = error.osColor,
+                    onPrimary = onPrimary.osColor,
+                    onSecondary = onSecondary.osColor,
+                    onBackground = onBackground.osColor,
+                    onSurface = onSurface.osColor,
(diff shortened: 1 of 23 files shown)


android/ui/src/main/java/com/trafi/ui/theme/Typography.kt
```diff
@@ -33,14 +28,14 @@ data class MaasTypography internal constructor(
...
         defaultFontFamily: FontFamily = Inter,
-        headingXXL: TextStyle = TypographyScale.headingXXL.toTextStyle(),
-        headingXL: TextStyle = TypographyScale.headingXL.toTextStyle(),
-        headingL: TextStyle = TypographyScale.headingL.toTextStyle(),
-        headingM: TextStyle = TypographyScale.headingM.toTextStyle(),
-        textL: TextStyle = TypographyScale.textL.toTextStyle(),
-        textM: TextStyle = TypographyScale.textM.toTextStyle(),
-        textS: TextStyle = TypographyScale.textS.toTextStyle(),
-        label: TextStyle = TypographyScale.label.toTextStyle(),
+        headingXXL: TextStyle = TypographyScale.headingXXL,
```
justasm, line 31: We avoid mapping on Android 🎉

android/ui/src/main/java/com/trafi/ui/theme/CurrentTheme.kt
```diff
@@ -0,0 +1,63 @@
...
+import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.unit.Dp
+import com.trafi.ui.theme.internal.CurrentColorPalette
+import com.trafi.ui.theme.internal.CurrentCornerRadiusScale
+import com.trafi.ui.theme.internal.CurrentSpacingScale
+import com.trafi.ui.theme.internal.CurrentTheme
+import com.trafi.ui.theme.internal.CurrentTypographyScale
+
+@Composable
+internal val currentTheme: CurrentTheme
```
justasm, line 14: Currently intended for internal use only

common/core/src/iosMain/kotlin/com/trafi/ui/theme/internal/OsTextStyle.kt
```diff
@@ -0,0 +1,46 @@
...
+    if (fontStyle?.os != null) {
+        descriptor = descriptor.fontDescriptorWithSymbolicTraits(fontStyle.os!!) ?: descriptor
+    }
+    if (fontWeight != null) {
+        descriptor = descriptor.fontDescriptorByAddingAttributes(
+            mapOf(UIFontDescriptorTraitsAttribute to mapOf(UIFontWeightTrait to fontWeight.os))
+        )
+    }
+    if (lineHeight != null) {
+        // TODO: No way to add line height to UIFont :(
```
justasm, line 23: Remains to be addressed in the future

ios/Sources/MaaSTheme/Typography.swift
```diff
@@ -1,186 +1,138 @@
 import MaasCore
 import SwiftUI
 
-public struct HeadingXXL: EnvironmentKey {
-    public static var defaultValue: Font {
-        TypographyScale().headingXXL.font
-    }
+struct HeadingXXL: EnvironmentKey {
+    static var defaultValue: UIFont { TypographyScale().headingXXL }
```
justasm, line 5: Less mapping on iOS 🎉

domasn (approved):

PR #39, opened 2020-12-10
Seamless theme types in Kotlin/Native

justasm:
To share theme-aware UI component values across iOS and Android, we can move them to Kotlin/Native. The initial implementation in #25 moved primitive constants to K/N, but required mapping between primitives and Compose/SwiftUI-native types. Since the mapping had only to be done once, this was not an issue. It becomes an issue when we declare constants for specific UI components (see e.g. #34) as the mapping then has to be done inside of each component, on both iOS & Android.

The goal is to make all theme types - `TextStyle`, `Color` & `Dp` - interop as seamlessly as possible with SwiftUI and Jetpack Compose types even when shared via Kotlin Multiplatform, avoiding mapping where possible. Seamless `TextStyle` was added in #37. This PR adds seamless usage of `Color` and `Dp`.

There are two challenges with `Color` and `Dp` -
1. Compose `TextStyle` is a simple (`data`) `class`, which allows us to use `expect class TextStyle` and map to any Obj-C class (e.g. `UIFont`). Compose `Color` and `Dp`, on the other hand, are `inline class`es, which forces us to use `expect inline class` for seamless interop between K/N & Compose. We cannot `actual typealias` `inline class`es to any Obj-C class.
2. K/N has interop with Obj-C only, not with Swift. SwiftUI types - `Font`, `Color` and `CGFloat` - are Swift types, which means we cannot avoid additional mapping on the Swift side regardless.

As a result, we choose to achieve seamless interop between Kotlin/Android & Kotlin/Native, where it is possible, and present a solution for painless mapping in Swift.

#### Android
Using expect / actual, all three types map directly to Jetpack Compose types by the same name using `actual typealias`. This means no type mapping is required between the Android and Kotlin/Native bridge.

#### iOS
All three types are seamlessly mapped through a `@dynamicMemberLookup struct Kotlin<T>`. For example, let's look at #34 Kotlin constants for button and their types when accessing from Swift:
```kotlin
class ButtonConstants(theme: CurrentTheme) {
     val defaultColor = theme.colorPalette.primary                       // UInt64
     val defaultContentColor = theme.colorPalette.onPrimary              // UInt64
     val textStyle = theme.typographyScale.textL.copy(fontWeight = Bold) // UIFont
     val cornerRadius = theme.cornerRadiusScale.buttonRadius             // Float
 }
```
Using a `Kotlin` wrapper perfectly handles the mapping to types preferred by SwiftUI using key paths:
```swift
var constants: Kotlin<ButtonConstants> { Kotlin(ButtonConstants(theme: currentTheme)) }

constants.defaultColor        // Color
constants.defaultContentColor // Color
constants.textStyle           // Font
constants.cornerRadius        // CGFloat
```

### DIFF
--- a/android/ui/src/main/java/com/trafi/ui/theme/Color.kt
+++ b/android/ui/src/main/java/com/trafi/ui/theme/Color.kt
@@ -9,29 +9,29 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.graphics.Color
 import com.trafi.ui.theme.internal.ColorPalette
 
-@Stable val Grey100 = Color(ColorPalette.Gray100)
-@Stable val Grey200 = Color(ColorPalette.Gray200)
-@Stable val Grey300 = Color(ColorPalette.Gray300)
-@Stable val Grey400 = Color(ColorPalette.Gray400)
-@Stable val Grey500 = Color(ColorPalette.Gray500)
-@Stable val Grey600 = Color(ColorPalette.Gray600)
-@Stable val Grey700 = Color(ColorPalette.Gray700)
-@Stable val Grey800 = Color(ColorPalette.Gray800)
-@Stable val Grey900 = Color(ColorPalette.Gray900)
+@Stable val Grey100 = ColorPalette.Gray100
+@Stable val Grey200 = ColorPalette.Gray200
+@Stable val Grey300 = ColorPalette.Gray300
+@Stable val Grey400 = ColorPalette.Gray400
+@Stable val Grey500 = ColorPalette.Gray500
+@Stable val Grey600 = ColorPalette.Gray600
+@Stable val Grey700 = ColorPalette.Gray700
+@Stable val Grey800 = ColorPalette.Gray800
+@Stable val Grey900 = ColorPalette.Gray900
 
 fun MaasTheme.lightColors(
-    primary: Color = Color(ColorPalette.DefaultLight.Primary),
-    primaryVariant: Color = Color(ColorPalette.DefaultLight.PrimaryVariant),
-    secondary: Color = Color(ColorPalette.DefaultLight.Secondary),
-    secondaryVariant: Color = Color(ColorPalette.DefaultLight.SecondaryVariant),
-    background: Color = Color(ColorPalette.DefaultLight.Background),
-    surface: Color = Color(ColorPalette.DefaultLight.Surface),
-    error: Color = Color(ColorPalette.DefaultLight.Error),
-    onPrimary: Color = Color(ColorPalette.DefaultLight.OnPrimary),
-    onSecondary: Color = Color(ColorPalette.DefaultLight.OnSecondary),
-    onBackground: Color = Color(ColorPalette.DefaultLight.OnBackground),
-    onSurface: Color = Color(ColorPalette.DefaultLight.OnSurface),
-    onError: Color = Color(ColorPalette.DefaultLight.OnError)
+    primary: Color = ColorPalette.DefaultLight.Primary,
+    primaryVariant: Color = ColorPalette.DefaultLight.PrimaryVariant,
+    secondary: Color = ColorPalette.DefaultLight.Secondary,
+    secondaryVariant: Color = ColorPalette.DefaultLight.SecondaryVariant,
+    background: Color = ColorPalette.DefaultLight.Background,
+    surface: Color = ColorPalette.DefaultLight.Surface,
+    error: Color = ColorPalette.DefaultLight.Error,
+    onPrimary: Color = ColorPalette.DefaultLight.OnPrimary,
+    onSecondary: Color = ColorPalette.DefaultLight.OnSecondary,
+    onBackground: Color = ColorPalette.DefaultLight.OnBackground,
+    onSurface: Color = ColorPalette.DefaultLight.OnSurface,
+    onError: Color = ColorPalette.DefaultLight.OnError,
 ): MaasColorPalette = MaasColorPalette(
     primary = primary,
     primaryVariant = primaryVariant,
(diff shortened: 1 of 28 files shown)


android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -45,22 +45,22 @@ fun MaasTheme.lightColors(
     onBackground = onBackground,
     onSurface = onSurface,
     onError = onError,
-    isLight = true
+    isLight = true,
```
eligijusTrafi, line 48:
```suggestion
    isLight = true
```

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -45,22 +45,22 @@ fun MaasTheme.lightColors(
...
+    secondary: Color = ColorPalette.DefaultDark.Secondary,
+    secondaryVariant: Color = ColorPalette.DefaultDark.SecondaryVariant,
+    background: Color = ColorPalette.DefaultDark.Background,
+    surface: Color = ColorPalette.DefaultDark.Surface,
+    error: Color = ColorPalette.DefaultDark.Error,
+    onPrimary: Color = ColorPalette.DefaultDark.OnPrimary,
+    onSecondary: Color = ColorPalette.DefaultDark.OnSecondary,
+    onBackground: Color = ColorPalette.DefaultDark.OnBackground,
+    onSurface: Color = ColorPalette.DefaultDark.OnSurface,
+    onError: Color = ColorPalette.DefaultDark.OnError,
```
eligijusTrafi, line 63:
```suggestion
    onError: Color = ColorPalette.DefaultDark.OnError
```

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -74,7 +74,7 @@ fun MaasTheme.darkColors(
     onBackground = onBackground,
     onSurface = onSurface,
     onError = onError,
-    isLight = false
+    isLight = false,
```
eligijusTrafi, line 77:
```suggestion
    isLight = false
```

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -91,7 +91,7 @@ class MaasColorPalette(
     onBackground: Color,
     onSurface: Color,
     onError: Color,
-    isLight: Boolean
+    isLight: Boolean,
```
eligijusTrafi, line 94:
```suggestion
    isLight: Boolean
```

android/ui/src/main/java/com/trafi/ui/theme/CurrentTheme.kt
```diff
@@ -12,52 +9,43 @@ import com.trafi.ui.theme.internal.CurrentTypographyScale
...
+                secondary = secondary,
+                secondaryVariant = secondaryVariant,
+                background = background,
+                surface = surface,
+                error = error,
+                onPrimary = onPrimary,
+                onSecondary = onSecondary,
+                onBackground = onBackground,
+                onSurface = onSurface,
+                onError = onError,
```
eligijusTrafi, line 26:
```suggestion
                onError = onError
```

android/ui/src/main/java/com/trafi/ui/theme/CurrentTheme.kt
```diff
@@ -12,52 +9,43 @@ import com.trafi.ui.theme.internal.CurrentTypographyScale
...
+        typographyScale = with(MaasTheme.typography) {
+            CurrentTypographyScale(
+                headingXXL = headingXXL,
+                headingXL = headingXL,
+                headingL = headingL,
+                headingM = headingM,
+                textL = textL,
+                textM = textM,
+                textS = textS,
+                label = label,
```
eligijusTrafi, line 38:
```suggestion
                label = label
```

common/core/src/androidMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
```diff
@@ -1,30 +1,29 @@
...
 
-actual typealias OsTextStyle = TextStyle
+actual typealias TextStyle = androidx.compose.ui.text.TextStyle
 
-internal actual fun OsTextStyle.copy(
+internal actual fun TextStyle.copy(
     fontStyle: BasicFontStyle?,
     fontWeight: BasicFontWeight?,
     fontSize: Int?,
     lineHeight: Int?,
```
eligijusTrafi, line 13:
```suggestion
    lineHeight: Int?
```

common/core/src/androidMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
```diff
@@ -1,30 +1,29 @@
...
     fontSize = fontSize?.sp ?: this.fontSize,
     lineHeight = lineHeight?.sp ?: this.lineHeight
 )
 
-internal actual fun OsTextStyle(
+internal actual fun TextStyle(
     fontStyle: BasicFontStyle,
     fontWeight: BasicFontWeight,
     fontSize: Int,
     lineHeight: Int,
```
eligijusTrafi, line 25:
```suggestion
    lineHeight: Int
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentColorPalette.kt
```diff
@@ -1,16 +1,16 @@
...
+    val secondary: Color,
+    val secondaryVariant: Color,
+    val background: Color,
+    val surface: Color,
+    val error: Color,
+    val onPrimary: Color,
+    val onSecondary: Color,
+    val onBackground: Color,
+    val onSurface: Color,
+    val onError: Color,
```
eligijusTrafi, line 15:
```suggestion
    val onError: Color
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentCornerRadiusScale.kt
```diff
@@ -1,5 +1,5 @@
 package com.trafi.ui.theme.internal
 
 class CurrentCornerRadiusScale(
-    val buttonRadius: OsDimension,
+    val buttonRadius: Dp,
```
eligijusTrafi, line 4:
```suggestion
    val buttonRadius: Dp
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentSpacingScale.kt
```diff
@@ -1,5 +1,5 @@
 package com.trafi.ui.theme.internal
 
 class CurrentSpacingScale(
-    val globalMargin: OsDimension,
+    val globalMargin: Dp,
```
eligijusTrafi, line 4:
```suggestion
    val globalMargin: Dp
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentTypographyScale.kt
```diff
@@ -1,12 +1,12 @@
...
-    val textS: OsTextStyle,
-    val label: OsTextStyle,
+    val headingXXL: TextStyle,
+    val headingXL: TextStyle,
+    val headingL: TextStyle,
+    val headingM: TextStyle,
+    val textL: TextStyle,
+    val textM: TextStyle,
+    val textS: TextStyle,
+    val label: TextStyle,
```
eligijusTrafi, line 11:
```suggestion
    val label: TextStyle
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/Dsl.kt
```diff
@@ -11,7 +11,7 @@ internal class TextStyleBuilder(
     var fontSize: Int? = null,
     var lineHeight: Int? = null,
```
eligijusTrafi, line 12:
```suggestion
    var lineHeight: Int? = null
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
```diff
@@ -1,17 +1,17 @@
...
 
-expect class OsTextStyle
+expect class TextStyle
 
-internal expect fun OsTextStyle.copy(
+internal expect fun TextStyle.copy(
     fontStyle: BasicFontStyle? = null,
     fontWeight: BasicFontWeight? = null,
     fontSize: Int? = null,
     lineHeight: Int? = null,
```
eligijusTrafi, line 9:
```suggestion
    lineHeight: Int? = null
```

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
```diff
@@ -1,17 +1,17 @@
...
-): OsTextStyle
+): TextStyle
 
-internal expect fun OsTextStyle(
+internal expect fun TextStyle(
     fontStyle: BasicFontStyle,
     fontWeight: BasicFontWeight,
     fontSize: Int,
     lineHeight: Int,
```
eligijusTrafi, line 16:
```suggestion
    lineHeight: Int
```

eligijusTrafi (approved):
👏

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
justasm: This was intentional 🙂 Kotlin 1.4 supports trailing parameter commas. The primary benefits are
- if we add a new parameter at the end, this line won't change in git
- easier to rearrange lines if you want to change the parameter order

PR #41, opened 2020-12-11
Introduce snapshot tests

tomasmarciulionis:


android/app/build.gradle
```diff
@@ -66,4 +73,8 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.2.0'
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.2.0'
     implementation 'androidx.activity:activity-ktx:1.1.0'
+    implementation 'junit:junit:4.13.1'
```
justasm, line 76: These could be `androidTestImplementation`, right?

tomasmarciulionis: Tried that but no. Not sure exactly why but when the test tries to make screenshot, app doesn't come to expected state ("CREATED").

absimas:
Please avoid such repetitive and unclear commits like. Our commit tree becomes pretty messy :disappointed: 

You can do that with interactive rebasing and squash/fixup commits together.

image

PR #42, opened 2020-12-11
Android CI changes

absimas:


justasm:
Seems reasonable! What happens when the PR changes only iOS code yet the Android build workflow check is required to merge PRs?

absimas:
Great question. Couldn't find a good way to do that and it seems that statuses are simply stuck like on this PR.

If this becomes a problem I guess we can disable check requirements and use common sense / require more reviews when merging.

Maybe even a github action that approves a PR when statuses it deems necessary have passed.

PR #43, opened 2020-12-11
Fix iOS colors mapping from Kotlin

mikasdo:
After updating color mapping for Jetpack Compose [Color](https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/Color#representation)
iOS color maping was broken.

fix by shifting UInt64 by 32 bits

snapshot tests passed ✅


justasm (approved):
Thanks. Oversight in #40

PR #44, opened 2020-12-14
Added Readme.md for button

yochees:


justasm:
Sharing the same thoughts as on Slack..
1. Would be great to embed a screenshot of the component from snapshot testing, e.g. [Button](./ios/Tests/MaasTests/__Snapshots__/SnapSpec/Button.2x.png).
1. The idea was to put all component constants in this folder, so to make sense of the READMEs this file should be named `Button.md`, or we should move to a separate folder per component.

justasm (commented):
Sharing the same thoughts as on Slack..
1. Would be great to embed a screenshot of the component from snapshot testing, e.g. [Button](./ios/Tests/MaasTests/__Snapshots__/SnapSpec/Button.2x.png).
1. The idea was to put all component constants in this folder, so to make sense of the READMEs this file should be named `Button.md`, or we should move to a separate folder per component.

yochees:
Renamet file, added image 🚀

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/Button.md
```diff
@@ -0,0 +1,19 @@
...
+**Sizes:** Small, Normal, Large.  
+**States:** Normal, Active, Disabled  
+**Behaviour:** onTap changes state to “Active”. If interaction takes longer than 300ms changes state to “inProgress” [TBD].  
+
+## Recommendations
+- As a rule of thumb, there should be only one primary button per screen. If you need more actions, please consider secondary, and/or tertiary buttons.
+- Keep text concise. Text longer than 22 symbols, might be truncated on smaller screens.
+
+## API
+See [ButtonConstants.kt](https://github.com/trafi/maas-components/blob/main/common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/ButtonConstants.kt)
```
justasm, line 16:
```suggestion
See [ButtonConstants.kt](ButtonConstants.kt)
```

justasm (approved):
Looks good!

Remaining thoughts
- may be nice to add "Button" to the readme text itself so it's clear what component we're talking about and it's not just in the filename 🙂
- would be nice to embed the constants directly in the readme, [I found this tool which we could run via Github Actions](https://github.com/zakhenry/embedme) - something to look forward to in the future 😉

PR #45, opened 2020-12-14
Color palette refactor

eligijusTrafi:
cc @mikasdo @justasm


mikasdo:
Need to look at how we expose GrayScale customisation for SDK users

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -117,6 +113,8 @@ class MaasColorPalette(
         private set
     var onError by mutableStateOf(onError)
         private set
+    var grayScale by mutableStateOf(grayScale)
```
justasm, line 116: What do you think about calling this `gray` and the various variants just `100`, `200`, etc? So we could write `MaasTheme.colors.gray.100` 🤔

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/Color.kt
```diff
@@ -11,4 +11,20 @@ import kotlin.jvm.JvmName
 
 expect inline class Color(val value: ULong)
 
+internal fun Color.alpha(value : Float): Color {
```
justasm, line 14: Very clear Kotlin function, super nice job @mikasdo 👏 Good visibility modifier, using `require` to verify inputs, clear variable names 💪

justasm (commented):
Looking good! Do you plan to change the existing palette to add the "warning" and "success" colors defined in Figma?

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
eligijusTrafi: 🔝 good idea 🤔

justasm: 😮 what if wrapped in backticks, e.g.
```
val `100` = ..
```
Not pretty though, so probably not the way to go

eligijusTrafi:
Note that `Warning` , `Success` and `Danger` colors cannot be changed by outsiders right now.

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -8,16 +8,7 @@ import androidx.compose.runtime.mutableStateOf
...
-@Stable val Grey100 = ColorPalette.Gray100
-@Stable val Grey200 = ColorPalette.Gray200
-@Stable val Grey300 = ColorPalette.Gray300
-@Stable val Grey400 = ColorPalette.Gray400
-@Stable val Grey500 = ColorPalette.Gray500
-@Stable val Grey600 = ColorPalette.Gray600
-@Stable val Grey700 = ColorPalette.Gray700
-@Stable val Grey800 = ColorPalette.Gray800
-@Stable val Grey900 = ColorPalette.Gray900
+import com.trafi.ui.theme.internal.GrayScale
```
justasm, line 11: If we're exposing `GrayScale` to our SDK users, it should be in the `com.trafi.ui.theme` package (not `internal`)

justasm (approved):
So the main change with this PR is that grayscale is theme-able by SDK users and is dark-theme aware.

Oh, and please ask designers to sync their color palette on the canvas and in the sidebar styles.
Screen Shot 2020-12-15 at 12 36 44

justasm (approved):

mikasdo:
@eligijusTrafi After some renamings and changes iOS was not updated
We should not merge to main before both platforms are successfuly build ✅

PR #47, opened 2020-12-14
Nearby transit UI components

TomasSkar:
### AssetBox
Introducing [AssetBox.swift](https://github.com/trafi/maas-components/blob/c17c5612b5ef279eed01d05be511783756668d69/ios/Sources/MaaS/Utils/AssetBox.swift) allows us at the moment show empty box instead of empty view when we're displaying images. Visuals are visible in the snapshots below.

### Filter
[Filter.swift](https://github.com/trafi/maas-components/blob/c17c5612b5ef279eed01d05be511783756668d69/ios/Sources/MaaS/UI/Components/Filter.swift) / [Filter.kt](https://github.com/trafi/maas-components/blob/c17c5612b5ef279eed01d05be511783756668d69/android/ui/src/main/java/com/trafi/ui/component/Filter.kt) defines filter logic for single and multi select.

Multi Select Filter Preview:
MultiSelectFilterPreview 2x
Single Select Filter Preview:
SingleSelectFilterPreview 2x

### NearbyTransitFilterItem

[NearbyTransitFilterItemView.swift](https://github.com/trafi/maas-components/blob/c17c5612b5ef279eed01d05be511783756668d69/ios/Sources/MaaS/UI/Features/NearbyTransit/Components/NearbyTransitFilterItemView.swift) / [NearbyTransitFilterItem.kt](https://github.com/trafi/maas-components/blob/c17c5612b5ef279eed01d05be511783756668d69/android/transit/ui/src/main/java/com/trafi/maas/transit/ui/NearbyTransitFilterItem.kt) a filter item that goes in to [NearbyTransitFilterView.swift](https://github.com/trafi/maas-components/blob/c17c5612b5ef279eed01d05be511783756668d69/ios/Sources/MaaS/UI/Features/NearbyTransit/Components/NearbyTransitFilterView.swift)

Item Preview:
NearbyTransitFilterItemView 2x
NearbyTransitFilterView Preview:
NearbyTransitFilterView 2x


common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/nearby.transit/NearbyTransitFilterItemConstants.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.trafi.ui.theme.internal.dp
+
+class NearbyTransitFilterItemConstants(theme: CurrentTheme) {
+    val disabledColor = ColorPalette.Gray200
+    val itemMinHeight = 48.dp
+    val itemMinWidth = 48.dp
+    val minHeight = 32.dp
+    val minWidth = 32.dp
+    val imageHeight = 24.dp
+    val imageWidth = 24.dp
```
Ignelios, line 14: Ahhh... Feels like we need better naming for these parameters...

`minHeight/Width` -> `contentMinHeight/Width` what do you think ? 🤔

common/core/src/commonMain/kotlin/com/trafi/core/HttpClient.kt
```diff
@@ -7,9 +7,14 @@ import io.ktor.client.features.json.serializer.KotlinxSerializer
 import io.ktor.client.request.header
 import io.ktor.http.ContentType
 import io.ktor.http.contentType
+import io.ktor.client.features.logging.*
 import kotlinx.serialization.json.Json
 
 internal fun httpClient(apiKey: String) = HttpClient {
+    install(Logging) {
+        logger = Logger.DEFAULT
```
justasm, line 15: Unrelated to this PR 😉

common/core/src/commonMain/kotlin/com/trafi/core/model/FilterItem.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param id
+ * @param name
+ * @param icon
+ * @param color
+ * @param accentColor
+ * @param transportsIds
+ */
+@Serializable
+data class FilterItem(
```
justasm, line 16: Have these been added to the platform-api contract? If no, we'll lose this file when we regenerate the models 😟

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentCornerRadiusScale.kt
```diff
@@ -2,4 +2,5 @@ package com.trafi.ui.theme.internal
 
 class CurrentCornerRadiusScale(
     val buttonRadius: Dp,
+    val nearbyTransitFilterItemRadius: Dp
```
justasm, line 5: See https://github.com/trafi/maas-components/pull/51#discussion_r543084043
> General rules of thumb -
> 1. do not put configurable values into these global theme values unless necessary
> 1. to judge if necessary, consider whether this is something we theme globally per rollout in our own whitelabel apps

justasm:
@TomasSkar what tasks remain for this PR to be considered done?

justasm (approved):
I've reviewed the PR in detail and implemented suggested changes, so LGTM 😉

android/transit/ui/src/main/java/com/trafi/maas/transit/ui/NearbyTransitFilterItem.kt
```diff
@@ -0,0 +1,164 @@
...
+                    .width(constants.imageWidth)
+                    .height(constants.imageHeight),
+                colorFilter = ColorFilter.tint(color = item.accentColor.toColor())
+            )
+        }
+    }
+}
+
+@DrawableRes
+private fun String.iconRes(): Int {
```
stonevyt, line 74: @justasm Maybe there is a possibility to move these mappings in some general module as ui or other to make it reachable in all modules because the same mapping is used in RoutesSegment composable. I guess there will be more similar functions in the future. What do you think?

justasm: Yeah, I agree. We can move it around and consolidate it but it doesn't solve the issue that we're bundling Berlin-specific transport icons. We'll address this in a more holistic way when we tackle image / icon loading in the SDK.

justasm: The mapping in `RouteSegment` maps from `Schedule.transportType` to icon; here we map from `FilterItem.icon`. It's a coincidence that the strings match 🙂 It's not clear if we want to load the same assets for route segment badges as we do for filter items.

stonevyt: Good point 👍

PR #50, opened 2020-12-14
Improve snapshot tests

tomasmarciulionis:
Tried to use dependency injection, but it seemed that it would be overhead for current state of SDK as no dependency injection is currently not used. So, just refactored tests to be more reusable.


android/app/src/androidTest/java/com/trafi/example/ButtonPreviews.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun init() {
+        previewTest = PreviewTest(composeTestRule = composeTestRule)
+    }
+
+    @Test
+    fun buttonTest() {
+        previewTest
+            .renderAndTestView {
+                MaasTheme(cornerRadius = MaasCornerRadius(buttonRadius = 0.dp)) {
+                    Button("Unlock now", onClick = {})
```
justasm, line 30: We could call the preview functions directly here, right? e.g. `ButtonStyledCornerRadiusPreview()`

justasm (approved):
Not clear how the Badge poster screenshot is related to this PR.

tomasmarciulionis:
> Not clear how the Badge poster screenshot is related to this PR. Also, as @absimas has previously mentioned, you can rebase and / or squash commits to clean up the git history and avoid commits like "WIP".

Removed unnecessary file.

android/app/src/androidTest/java/com/trafi/example/ButtonPreviews.kt
tomasmarciulionis: We should be able to call preview functions directly.

PR #51, opened 2020-12-14
Define badge constants

stonevyt:


android/ui/src/main/java/com/trafi/ui/theme/Spacing.kt
```diff
@@ -21,5 +21,10 @@ object Spacing {
 
 @Immutable
 data class MaasSpacing(
-    val globalMargin: Dp = SpacingScale.Default.GlobalMargin
+    val globalMargin: Dp = SpacingScale.Default.GlobalMargin,
+    val badgeSpacer: Dp = SpacingScale.xxs,
```
mikasdo, line 25: I think badge spacer should only be part of BadgeConstants. 
MaasSpacing should not know about any component 🤔
Or I missed something?

justasm: Yes, I completely agree 👍

justasm: General rules of thumb -
1. do not put configurable values into these global theme values unless necessary
1. to judge if necessary, consider whether this is something we theme globally per rollout in our own whitelabel apps

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/BadgeConstants.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    val spacer = theme.spacingScale.badgeSpacer
+
+    val horizontalPaddingMedium = theme.spacingScale.horizontalPaddingMedium
+    val horizontalPaddingSmall = theme.spacingScale.horizontalPaddingSmall
+
+    val verticalPaddingMedium = theme.spacingScale.verticalPaddingMedium
+    val verticalPaddingSmall = theme.spacingScale.verticalPaddingSmall
+
+    val border = (1.5f).dp
```
justasm, line 36: I'd suggest adding an `internal val Double.dp get() = Dp(toFloat())` property to our common code so this can be written as `1.5.dp`, would be less syntax for designers to worry about.

justasm: Also, border what? `borderWidth` perhaps?

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/BadgeConstants.kt
```diff
@@ -0,0 +1,37 @@
...
+    val disabledColor = Gray300
+
+    val textStyleMedium = theme.typographyScale.textM.copy(fontWeight = BasicFontWeight.Bold)
+    val textStyleSmall = theme.typographyScale.textS.copy(fontWeight = BasicFontWeight.Bold)
+
+    val minHeightMedium = 24.dp
+    val minHeightSmall = 16.dp
+
+    val cornerRadiusMedium = theme.cornerRadiusScale.mediumBadgeRadius
+    val cornerRadiusSmall = theme.cornerRadiusScale.smallBadgeRadius
```
justasm, line 20: If you agree that we don't want to let SDK users theme these globally, just inline the values here, i.e. write `cornerRadiusMedium = CornerRadiusScale.xs`

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentSpacingScale.kt
```diff
@@ -2,4 +2,12 @@ package com.trafi.ui.theme.internal
 
 class CurrentSpacingScale(
     val globalMargin: Dp,
+
+    val badgeSpacer: Dp,
+
+    val horizontalPaddingMedium: Dp,
+    val horizontalPaddingSmall: Dp,
```
justasm, line 9: Do we expect to have more "medium" and "small" components than badges? Should they all share these same constants? If no, these probably shouldn't be globally theme-able. If yes, we should consider using a "shape" concept like material design uses, where they let SDK users globally override small, medium and large "shapes", where each shape can have its own properties like corner radius, padding, etc.

justasm (changes requested):

android/ui/src/main/java/com/trafi/ui/theme/CornerRadius.kt
```diff
@@ -23,5 +23,7 @@ internal val Dp.isRound: Boolean
 
 @Immutable
 data class MaasCornerRadius(
-    val buttonRadius: Dp = CornerRadiusScale.Default.ButtonRadius
+    val buttonRadius: Dp = CornerRadiusScale.Default.ButtonRadius,
+    val smallBadgeRadius: Dp = CornerRadiusScale.xxs,
+    val mediumBadgeRadius: Dp = CornerRadiusScale.xs
```
justasm: Don't forget to remove these, too 🙂

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/BadgeConstants.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.trafi.ui.theme.internal.*
+import com.trafi.ui.theme.internal.ColorPalette.Gray300
+import com.trafi.ui.theme.internal.copy
+import com.trafi.ui.theme.internal.dp
+
+class BadgeConstants(theme: CurrentTheme) {
+    val defaultContentColor = theme.colorPalette.onPrimary
+    val disabledColor = Gray300
+
+    val textStyleMedium = theme.typographyScale.textM.copy(fontWeight = BasicFontWeight.Bold)
```
justasm, line 12: Can import `BasicFontWeight.*` to be able to just write `Bold`

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/CurrentCornerRadiusScale.kt
```diff
@@ -1,5 +1,5 @@
 package com.trafi.ui.theme.internal
 
 class CurrentCornerRadiusScale(
-    val buttonRadius: Dp,
+    val buttonRadius: Dp
```
justasm, line 4: Leave these commas please, will lead to more accurate diffs in the future 🙂

justasm (approved):
Great 👏

PR #54, opened 2020-12-15
Add color field to TextStyle

eligijusTrafi:


common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
```diff
@@ -7,11 +7,13 @@ internal expect fun TextStyle.copy(
     fontWeight: BasicFontWeight? = null,
     fontSize: Int? = null,
     lineHeight: Int? = null,
+    color: Color? = null
```
justasm, line 10: Don't forget the trailing commas 🙂

justasm (changes requested):
Need to add the constructor and copy function implementations for `iosMain`

justasm (approved):
We can add the iOS implementation separately

PR #55, opened 2020-12-15
Fix ios fonts

mikasdo:


common/core/src/iosMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
```diff
@@ -30,7 +30,7 @@ internal actual fun TextStyle(
     fontWeight: BasicFontWeight,
     fontSize: Int,
     lineHeight: Int,
-): TextStyle = UIFont.systemFontOfSize(fontSize.os).copy(fontStyle, fontWeight, lineHeight)
+): TextStyle = UIFont.systemFontOfSize(fontSize.os).copy(fontStyle, fontWeight, fontSize, lineHeight)
```
justasm, line 33: 🙈 good catch! I'd recommend using named parameters to avoid similar issues in the future, i.e.

```kotlin
UIFont.systemFontOfSize(fontSize.os).copy(
  fontStyle = fontStyle,
  fontWeight = fontWeight,
  lineHeight = lineHeight,
)
```
(assuming `fontSize` doesn't need to be modified a second time)

justasm (approved):

common/core/src/iosMain/kotlin/com/trafi/ui/theme/internal/TextStyle.kt
Ignelios: Sure!

Ignelios (approved):

PR #57, opened 2020-12-15
Make Badge component

stonevyt:


android/routes/ui/src/main/java/com/trafi/routes/ui/RouteSegment.kt
```diff
@@ -146,59 +138,111 @@ fun RouteSegment(segment: RouteSegment, modifier: Modifier = Modifier) {
...
+private fun BadgeWithoutIconPreview() {
+    Badge(
+        badge = BadgeInfo("5G", Color.Magenta),
+        badgeType = BadgeType.MEDIUM_BADGE
+    )
 }
+
+@Preview
+@Composable
+private fun StackedBadgePreview() {
```
justasm, line 216: Seems like these previews belong next to the `Badge` component 🙂

android/routes/ui/src/main/java/com/trafi/routes/ui/RouteSegment.kt
```diff
@@ -146,59 +138,111 @@ fun RouteSegment(segment: RouteSegment, modifier: Modifier = Modifier) {
...
+        badgeType = BadgeType.MEDIUM_BADGE
+    )
+}
+
+//@Preview(showBackground = false)
+//@Composable
+//private fun RouteSegmentPreview(
+//    @PreviewParameter(RouteSegmentPreviewParameterProvider::class) segment: RouteSegment
+//) {
+//    RouteSegment(segment)
```
justasm, line 247: We probably want to leave the preview for `RouteSegment` available.

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+private fun getBadgeTextStyle(type: BadgeType): TextStyle {
+    return when (type) {
+        SMALL_BADGE -> constants.textStyleSmall
+        MEDIUM_BADGE -> constants.textStyleMedium
+    }
+}
+
+enum class BadgeType {
+    SMALL_BADGE,
+    MEDIUM_BADGE
```
justasm, line 257: I'd recommend using a slightly more readable naming convention 
```kotlin
enum class BadgeType {
  Small,
  Medium,
}
```
1. avoid repeating "badge"
1. avoid upper snake case for closer consistency with Swift
1. trailing commas 🎉

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+@Composable
+private fun getBadgeVerticalPadding(type: BadgeType): Dp {
+    return when (type) {
+        SMALL_BADGE -> 0.dp
+        MEDIUM_BADGE -> constants.verticalPaddingMedium
+    }
+}
+
+@Composable
+private fun getBadgeTextStyle(type: BadgeType): TextStyle {
```
justasm, line 248: More idiomatic in Kotlin to write this (and most other derived values from `BadgeType`) as an extension property (or extension function if additional parameters are needed), i.e.

```kotlin
private val BadgeType.textStyle: TextStyle get() = //..
```

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+                colorFilter = ColorFilter.tint(AmbientContentColor.current),
+                modifier = Modifier.align(Alignment.CenterVertically)
+                    .width(constants.iconWidth)
+                    .height(if (isHiddenLayoutFiller) 0.dp else constants.iconHeight)
+            )
+        }
+        text?.let {
+            Text(
+                text = text,
+                style = getBadgeTextStyle(badgeType).copy(fontWeight = FontWeight.Bold),
```
justasm, line 206: `fontWeight` should be set in KMM for consistency with iOS

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+                modifier = Modifier.align(Alignment.CenterVertically)
+                    .padding(bottom = 1.dp)
+                    .wrapContentHeight(unbounded = true)
```
justasm, line 209: 😱 what do these lines do? Seems hacky, but if it works, no worries 🙂

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+}
+
+@Composable
+private fun BadgeFiller(
+    badgeType: BadgeType,
+    vector: ImageVector?,
+    text: String?,
+    isHiddenLayoutFiller: Boolean = false,
```
justasm, line 184: Optional parameters with default values should be declared after mandatory parameters.

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+            )
+        }
+    }
+}
+
+
+@OptIn(ExperimentalLayout::class)
+@Composable
+private fun StackedBadge(
+    modifier: Modifier = Modifier,
```
justasm, line 115: Compose convention is that `Modifier` is the first optional parameter (after all required parameters)

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+    badge: BadgeInfo?,
+    alternativeBadges: List<BadgeInfo>,
+    vector: ImageVector?,
+    subbadge: ImageVector? = null,
+    badgeType: BadgeType
+) {
+    ConstraintLayout {
+        val altBadgesSublist =
+            if (alternativeBadges.size > 2) alternativeBadges.subList(0, 2) else alternativeBadges
```
justasm, line 124: Could be slightly simpler `alternativeBadges.take(2)`

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+        ),
+        horizontalArrangement = Arrangement.spacedBy(constants.spacer)
+    ) {
+        vector?.let {
+            Image(
+                vector,
+                colorFilter = ColorFilter.tint(AmbientContentColor.current),
+                modifier = Modifier.align(Alignment.CenterVertically)
+                    .width(constants.iconWidth)
+                    .height(if (isHiddenLayoutFiller) 0.dp else constants.iconHeight)
```
justasm, line 200: Perhaps `Image` shouldn't be added at all if `isHiddenLayoutFiller == true`? Could replace the whole block with some spacer, e.g. `Box`. Would avoid unnecessarily loading the vector image.

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+data class BadgeInfo(
+    val text: String?,
+    val color: Color,
+    val textColor: Color? = null
+)
+
+@OptIn(ExperimentalLayout::class)
+@Composable
+fun Badge(
+    modifier: Modifier = Modifier,
```
justasm, line 38: Same comment re Compose convention for `Modifier`

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+    badge: BadgeInfo?,
```
justasm, line 39: Does it make sense for this to be nullable? Could it be required?

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+    alternativeBadges: List<BadgeInfo> = listOf(),
+    vector: ImageVector? = null,
+    subbadge: ImageVector? = null,
```
justasm, line 42: Slightly misleading name, using the term "badge" but it's an icon / image 🤔

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,258 @@
...
+    badgeType: BadgeType,
```
justasm, line 43: Could this have a reasonable default value, e.g. `Medium`? Also interesting why you decided to split `BadgeType` from `BadgeInfo`.

justasm (changes requested):

android/ui/src/main/java/com/trafi/ui/theme/Color.kt
```diff
@@ -136,3 +136,6 @@ class MaasColorPalette(
         isLight = other.isLight
     }
 }
+
+fun String.parseColor(): Color =
+    Color(android.graphics.Color.parseColor("#$this"))
```
justasm, line 141: We'd discussed with @TomasSkar that we want to eventually move this to Kotlin/Native so that iOS could use the same function, too 🙂

TomasSkar: https://github.com/trafi/maas-components/pull/64

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,354 @@
...
+        Medium -> constants.textStyleMedium
+    }
+
+enum class BadgeType {
+    Small,
+    Medium,
+}
+
+private fun Dp.plusIf(shouldInclude: Boolean, other: Dp): Dp {
+    return if (shouldInclude) Dp(value = this.value + other.value) else Dp(value = this.value)
```
justasm, line 266: I'm pretty sure `Dp` defines `operator fun plus` so we should be able to write `if (shouldInclude) this + other else this`

android/ui/src/main/java/com/trafi/ui/Badge.kt
```diff
@@ -0,0 +1,354 @@
...
+                    isStackedBadge = true
+                )
+            }
+        }
+        Surface(
+            color = badge.backgroundColor,
+            contentColor = badge.contentColor ?: constants.defaultContentColor,
+            border = BorderStroke(constants.borderWidth, color = constants.defaultContentColor),
+            shape = RoundedCornerShape(badgeType.badgeRounding),
+            modifier = modifier
```
justasm, line 147: Not clear if this is correct, typically the incoming modifier should be applied to the root / parent layout (i.e. `ConstraintLayout` in this case), but perhaps it works out in this case.

justasm (approved):
Nice refactor 👍 Our third Android component, congrats 🎉

PR #63, opened 2020-12-16
Enable explicit-api mode for Android Kotlin modules

justasm:
[Read more in the official docs](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors).

> In this mode, the compiler performs additional checks that help make the library’s API clearer and more consistent. It adds the following requirements for declarations exposed to the library’s public API:
> - Visibility modifiers are required for declarations if the default visibility exposes them to the public API. This helps ensure that no declarations are exposed to the public API unintentionally.
> - Explicit type specifications are required for properties and functions that are exposed to the public API. This guarantees that API users are aware of the types of API members they use.

Eventually want to enable this for the `:common:core` module, too, however we need to change the API model generation there first.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -40,13 +40,17 @@ subprojects {
 
 subprojects { project ->
     afterEvaluate {
-        tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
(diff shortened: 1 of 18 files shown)


eligijusTrafi (approved):
👮

PR #66, opened 2020-12-16
[Badge]

ignasvalotka:
image


ios/Sources/MaaS/UI/Components/Badge.swift
```diff
@@ -1,65 +1,309 @@
...
+
     public struct InputType {
-        let color: Color?
+        let type: BadgeType
+        let backgroundColor: Color?
+        let foregroundColor: Color?
         let icon: UIImage?
         let text: String?
+        let alternativeBadgesColors: [Color]
+        let subBadge: UIImage?
```
justasm, line 22: @stonevyt called this `subbadgeIcon` in Compose, would be great to strive for consistency 🌟

justasm (commented):
Looks good overall to me, would be nice to hear some input from iOS colleagues regarding SwiftUI 🙂

ios/Sources/MaaS/UI/Components/Badge.swift
ignasvalotka: Yes! missed this one

ios/Sources/MaaS/UI/Components/Badge.swift
```diff
@@ -1,65 +1,309 @@
...
+            }
+        }
+        .fixedSize()
+        .subBadge(image: input.base.subBadgeIcon,
+                  width: constants.subBadgeIconWidth,
+                  height: constants.subBadgeIconHeight)
+    }
+
+    private var badges: [Badge.InputType] {
+        let maxAlternativeBadges = 2
```
mikasdo, line 223: This should be a shared constant

ios/Sources/MaaS/UI/Components/Badge.swift
```diff
@@ -1,65 +1,309 @@
...
+        let alternativeBadges = Array(input.colors.prefix(maxAlternativeBadges)).toAlternativeBadges()
+        return alternativeBadges + [input.base]
+    }
+
+    private func spacing(index: Int) -> CGFloat {
+        CGFloat(2.5 * Double(index))
```
mikasdo, line 229: This should probably also be a constant

ignasvalotka: Yes! This actually can be computed from the constants we already have!

ios/Sources/MaaS/UI/Components/Badge.swift
ignasvalotka: Agreed

common/core/src/commonMain/kotlin/com/trafi/ui/theme/internal/Color.kt
```diff
@@ -28,12 +28,3 @@ internal fun Color.alpha(value : Float): Color {
 }
 
 internal fun Color(color: Long) = Color((color.toULong() and 0xffffffffUL) shl 32)
-
-fun String?.toColor(fallback: Long = 0xFF_000000): Color {
```
justasm, line 32: This shouldn't be removed

mikasdo (approved):

PR #67, opened 2020-12-16
Artifacts for iOS test action

aistetests:


justasm (approved):
If you squash the commits with the PR title as its message, it would result in a much cleaner git history 🙏

PR #68, opened 2020-12-16
Increased snapshot verification accuracy

aistetests:
Discovered that Badge test still passes if I change text from 12 to 1ą or even 11ą.
It only fails if I change it to 111ą.

Played a bit with snapshot accuracy and from 0.996 and up the Button test starts to fail because of minor text shifts in multiple previews. With 0.995 Button test starts to pass and Badge test fails if text is changed to 11ą (still passes if changed to 1ą).

So keeping the 0.995 setting and this means tests may miss a change of one character.

### DIFF
--- a/ios/Tests/MaasTests/Components.swift
+++ b/ios/Tests/MaasTests/Components.swift
@@ -3,8 +3,8 @@ import Quick
 
 final class Components: QuickSpec {
     override func spec() {
-        snapshot(Badge_Previews.self)
-        snapshot(Button_Previews.self)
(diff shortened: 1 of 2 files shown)


justasm (approved):
IIRC we discussed yesterday that we could parametrize this and specify a less strict default for only those components that need it.

justasm (approved):
Nice!

aistetests:
@justasm yeah, we did! Just pushed parameterized version, unfortunately Badge also needed precision <1, but was OK with stricter precision than Button. It seems we will need to see with time how different precision we'll need for different components.

PR #69, opened 2020-12-16
Create android_tests.yml

tomasmarciulionis:
Adding github action to run android tests


.github/workflows/android_tests.yml
```diff
@@ -0,0 +1,38 @@
...
+# Controls when the action will run. 
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
```
justasm, line 14:
```suggestion
  test:
```

.github/workflows/android_tests.yml
```diff
@@ -0,0 +1,43 @@
...
+         with:
+           api-level: 28
+           target: default
+           arch: x86
+           profile: Nexus 6
+           disable-animations: true
+           script: ./gradlew executeScreenshotTests
+           working-directory: ./android
+      
+      -  name: Publish artefacts
```
justasm, line 35:
```suggestion
      -  name: Publish artifacts
```

.github/workflows/android_tests.yml
```diff
@@ -0,0 +1,43 @@
...
+         if: always()
+         uses: actions/upload-artifact@v2
+         with:
+          name: tests_results.html
+          path: /Users/runner/work/maas-components/maas-components/android/app/build/reports/shot/verification/index.html
```
justasm, line 40: Is the html file enough? Having downloaded it, it does not contain the actual images and diffs 😞

.github/workflows/android_tests.yml
```diff
@@ -0,0 +1,43 @@
...
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test:
+    runs-on: macos-latest
+    timeout-minutes: 30
```
justasm, line 16: We should aim for 10min 🚀

.github/workflows/android_tests.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test:
+    runs-on: macos-latest
```
justasm, line 15: Do we need to run on a macos host? Perhaps `ubuntu-latest` would suffice? My assumption is that macOS hosts may be slower / more expensive, but not sure if that's true.

justasm (approved):

PR #76, opened 2020-12-17
Generate docs using Sphinx

justasm:
Generate docs using [Sphinx](https://www.sphinx-doc.org/en/master/).

By default, Sphinx uses reStructuredText, however, we want to use Markdown files. To that end, the [MyST parser](https://myst-parser.readthedocs.io/) allows authoring and parsing Markdown files. We can use rst commands in md documents parsed by MyST, but should try to minimize this to make our documentation more portable (in case we want to switch out Sphinx for another static site generator).

The [Furo theme](https://pradyunsg.me/furo/) is used as something simple to get us started.

All design documentation has been moved to the `docs/design/` folder.

View the documentation site at https://trafi.github.io/maas-components/


Ignelios (approved):
Generated docs looking good! Noticed that Figma preview don't have a loading indicator...

justasm:
Yeah, not sure if we can do much about that loading indicator 😕 We're using [Figma's official embed functionality](https://www.figma.com/developers/embed).

PR #77, opened 2020-12-18
Info button

mikasdo:


android/app/src/androidTest/java/com/trafi/example/TestPreviews.kt
```diff
@@ -1,11 +1,14 @@
...
-import com.trafi.ui.ButtonPreview
-import com.trafi.ui.ButtonStyledCornerRadiusPreview
+import com.trafi.ui.*
 
 @Suppress("RemoveExplicitTypeArguments")
 val testPreviews = listOf<Pair<@Composable () -> Unit, String>>(
         Pair({ ButtonPreview() }, "ButtonPreview"),
-        Pair({ ButtonStyledCornerRadiusPreview() }, "ButtonStyledCornerRadiusPreview")
+        Pair({ ButtonStyledCornerRadiusPreview() }, "ButtonStyledCornerRadiusPreview"),
+        Pair({ InfoButtonLongTextPreview() }, "InfoButtonLongTextPreview"),
```
justasm, line 10: I wonder if there is value in using the function name as the test name 🤔 See iOS using titles like "Long title" and "Themed" in `InfoButton.swift`. Ideally this naming would even match so we could compare apples to apples across Android & iOS 🌟

android/ui/src/main/java/com/trafi/ui/InfoButton.kt
```diff
@@ -0,0 +1,156 @@
...
+@Composable
+private val constants
+    get() = InfoButtonConstants(currentTheme)
+
+@Composable
+fun InfoButton(
+    text: String,
+    modifier: Modifier = Modifier,
+    @DrawableRes iconRes: Int = R.drawable.ic_help_info_s,
+    onClick: () -> Unit,
```
justasm, line 34: Required parameters should come before optional ones

ios/Sources/MaaS/UI/Components/InfoButton.swift
```diff
@@ -0,0 +1,127 @@
+import SwiftUI
+import MaasTheme
+
+struct InfoButton: View {
+    
+    public struct InputType {
+        public let text: String
+        public let icon: Image
+        public let foreground: Color?
```
justasm, line 9: Is "foreground" a commonly used term in SwiftUI? Asking because the Android implementation uses the term "content color" - it's fine to diverge if it more closely matches platform conventions, but if not, we should strive for consistency.

justasm (approved):
Nice work 👏

Expressive shared gradients 💪 Unfortunately, seems like it's not used in the Compose implementation 😞 Did you have plans to support gradients in Compose, or decided not to do it?

ios/Sources/MaaS/UI/Components/InfoButton.swift
mikasdo: ForegroundColor is used in SwiftUI, InfoButtonConstants Have been updated multiple times after this input was declared :D will update to "contentColor" to have consistency in all MaasComponents

mikasdo:
For now Gradient was used for a highlighted button in SwiftUI, Android uses default ripple effect, Gradient in Android will need to be implemented, when it's used somewhere else :)

ios/Sources/MaaS/UI/Components/InfoButton.swift
justasm: Let's have a discussion next year where it is more important to stick with consistent naming with SwiftUI / Compose respectively vs our own internal naming 🙂 I do think it's valuable to use naming expected by SwiftUI / Compose conventions, especially for components that we expect SDK users to be using directly.

android/app/src/androidTest/java/com/trafi/example/TestPreviews.kt
vbiv: I updated them to be more similar to iOS, but cannot make the same, because in that case we would have same name "Themed" for all the components and in the final test report there would be no way to distinguish them. Now we have one report for all tests, but we are planning to split it by components in the future, then names could be updated too. This is not the final setup yet :)

PR #82, opened 2020-12-21
Route search tab component

stonevyt:


android/routes/ui/src/main/java/com/trafi/routes/internal/LocationDisplayText.kt
```diff
@@ -1,4 +1,4 @@
-package com.trafi.routes.ui.internal
+package com.trafi.routes.internal
```
justasm, line 1: Could you please clarify why you decided to change the package of these files? They are directly related to routes UI.

android/routes/ui/src/main/java/com/trafi/routes/internal/RoutesViewModel.kt
```diff
@@ -11,7 +11,7 @@ import com.trafi.core.model.RoutesResult
 import com.trafi.routes.RoutesApi
 import kotlinx.coroutines.launch
 
-internal class RoutesViewModel(private val routesApi: RoutesApi) : ViewModel() {
+class RoutesViewModel(private val routesApi: RoutesApi) : ViewModel() {
```
justasm, line 14: This class should remain `internal` for now.

android/routes/ui/src/main/java/com/trafi/routes/ui/RoutesScreen.kt
```diff
@@ -143,34 +145,41 @@ fun RoutesScreen(
...
-                "No results",
-                modifier = modifier.align(Alignment.CenterHorizontally),
-                style = MaasTheme.typography.textL
-            )
-        }
-        RoutesResultState.Loading -> {
-            CircularProgressIndicator(
-                modifier = modifier.align(Alignment.CenterHorizontally)
-            )
+    Column {
```
justasm, line 161: Since `RouteSearchBody` is already called inside of a `Column` (and hence why it was an extension on `ColumnScope`), there should be no need to wrap it in another `Column` - just add the `Divider` as you did below. If a `Column` is necessary, then the root composable should respect the `modifier` parameter.

android/routes/ui/src/main/java/com/trafi/routes/ui/RoutesScreen.kt
```diff
@@ -198,6 +207,39 @@ private fun ColumnScope.LocationSearchBody(
     }
 }
 
+@Composable
+fun RouteSearchTabsList(
+    modifier: Modifier,
```
justasm, line 212: The `modifier: Modifier` parameter is always optional by convention.

android/routes/ui/src/main/java/com/trafi/routes/ui/RoutesScreen.kt
```diff
@@ -198,6 +207,39 @@ private fun ColumnScope.LocationSearchBody(
     }
 }
 
+@Composable
+fun RouteSearchTabsList(
```
justasm, line 211: This can be `private` for now as we don't expect SDK users to use this outside of `RoutesScreen()`

android/routes/ui/src/main/java/com/trafi/routes/ui/RoutesScreen.kt
```diff
@@ -198,6 +207,39 @@ private fun ColumnScope.LocationSearchBody(
...
+    modifier: Modifier,
+    vehicleTypes: List<TabItem>,
+    selectedItemId: String,
+    onVehicleClick: (String) -> Unit,
+    itemModifier: Modifier = Modifier,
```
justasm, line 216: Will we need to customize items from the outside? Seems like this parameter could be omitted.

android/routes/ui/src/main/java/com/trafi/routes/ui/RoutesScreen.kt
```diff
@@ -198,6 +207,39 @@ private fun ColumnScope.LocationSearchBody(
     }
 }
 
+@Composable
+fun RouteSearchTabsList(
+    modifier: Modifier,
+    vehicleTypes: List<TabItem>,
+    selectedItemId: String,
+    onVehicleClick: (String) -> Unit,
```
justasm, line 215: Naming -- `onRouteTabClick` or `onTabClick` would be clearer

android/routes/ui/src/main/java/com/trafi/routes/ui/RoutesScreen.kt
```diff
@@ -198,6 +207,39 @@ private fun ColumnScope.LocationSearchBody(
     }
 }
 
+@Composable
+fun RouteSearchTabsList(
+    modifier: Modifier,
+    vehicleTypes: List<TabItem>,
```
justasm, line 213: Perhaps `items` may be a better name?

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
```diff
@@ -0,0 +1,152 @@
...
+    val icon: String,
+    val active: Boolean,
+    val duration: String?,
+    val price: String?,
+)
+
+@Composable
+fun RouteSearchTab(
+    tabItem: TabItem,
+    modifier: Modifier,
```
justasm, line 35: `Modifier` should be optional.

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
```diff
@@ -0,0 +1,152 @@
...
+    isTabSelected: Boolean,
```
justasm, line 36: `isSelected` would be a simpler name 🙂 We're already in the context of "route search tab"

justasm (commented):
Nice use of shared constants 👏

android/routes/ui/src/main/java/com/trafi/routes/internal/LocationDisplayText.kt
stonevyt: @justasm I just thought that file in internal folder like ViewModel type clases are not directly related with ui, but if it is, then okay, I will leave it

justasm: I'd leave them there 🙂 They are related to UI, just not meant for SDK users to use directly.

android/routes/ui/src/main/java/com/trafi/routes/ui/internal/RoutesViewModel.kt
```diff
@@ -28,7 +28,7 @@ internal class RoutesViewModel(private val routesApi: RoutesApi) : ViewModel() {
     }
 }
 
-internal sealed class RoutesResultState {
+sealed class RoutesResultState {
```
justasm, line 31: Not clear if this change is meaningful 🙂

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
```diff
@@ -0,0 +1,151 @@
...
+    val duration: String?,
+    val price: String?,
+)
+
+@Composable
+fun RouteSearchTab(
+    tabItem: TabItem,
+    isSelected: Boolean,
+    modifier: Modifier = Modifier,
+    onRouteTabClick: () -> Unit,
```
justasm, line 36: Required parameters should be ordered before optional parameters

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
```diff
@@ -0,0 +1,151 @@
...
+            color = if (isSelected) constants.borderColorActive else constants.defaultBackgroundColor),
+        modifier = modifier
+            .padding(constants.padding)
+            .widthIn(min = constants.minTabWidth)
+            .heightIn(min = constants.minTabHeight)
+    ) {
+        Box(
+            modifier = modifier
+                .clickable(onClick = onRouteTabClick)
+                .background(if (isSelected) constants.activeBackgroundColor else constants.defaultBackgroundColor)
```
justasm, line 50: Could this background and `Modifier.clickable` be set on the parent `Card`? It supports a `backgroundColor` parameter.

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
```diff
@@ -0,0 +1,151 @@
...
+        shape = RoundedCornerShape(constants.borderRadius),
+        border = BorderStroke(constants.borderWidth,
+            color = if (isSelected) constants.borderColorActive else constants.defaultBackgroundColor),
+        modifier = modifier
+            .padding(constants.padding)
+            .widthIn(min = constants.minTabWidth)
+            .heightIn(min = constants.minTabHeight)
+    ) {
+        Box(
+            modifier = modifier
```
justasm, line 48: The incoming `modifier` is applied twice - for the `Card` and again for the `Box`

android/routes/ui/src/main/java/com/trafi/routes/ui/internal/RoutesViewModel.kt
stonevyt: @justasm Sorry, it left from testing

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
stonevyt: @justasm I just thought that click listeners goes at the end

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
stonevyt: @justasm This decision was made not accidentally because, when I tested with modifiers it needed same parameters for both layouts, but now I realised that it's not good, because with some attributes it will broke the layout, so I will fix this

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
stonevyt: @justasm First idea was the same as you described but it does not work like that:

- The problem with clickable is that when you set on card, the clickable effect goes off the card (because card has RounderCornerShape)
- With background problem is with active color, because there is primary color with alpha and when you set it on parent layout (in this case Card) the background looks weird

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
justasm: No, generally optional parameters go at the end. By Compose convention, `modifier: Modifier = Modifier` is the first optional parameter.

android/ui/src/main/java/com/trafi/ui/RouteSearchTab.kt
justasm: Got it

justasm (approved):
Nice work 👏

PR #85, opened 2021-01-05
Workflow to record Android test screenshots demo

justasm:
A demonstration of #84 at work.


justasm:
This was for demo purposes only.

PR #89, opened 2021-01-15
MaasCore

Ignelios:
Now we can rebuild a `MaasCore.xcframework` with **cmd+b** and we have a cached **MaasCore** it take a few seconds only 🎊 .

Next steps - try to use a AppleScript to have a progress bar or more details about MaasCore construction.

Just need to edit **Maas** scheme **Build/Pre-Actions** run script with this.
```
cd ~/(path to source)/common
./gradlew xcframeworkSimulator
```

No `.executable(...)` used.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 
 // MARK: - Setup
 
-private let environment: Environment = .development
+private let environment: Environment = .production
 
 let package = Package(
     name: "Maas",
--- a/common/core/build.gradle.kts
+++ b/common/core/build.gradle.kts
@@ -88,7 +88,7 @@ android {
 
 apply(from = "../../android/scripts/maven-meta.gradle")
 
-val xcframeworkPath = "build/bin/xcframework/MaasCore.xcframework"
+val xcframeworkPath = "../../ios/MaasCore/Sources/MaasCore/Core.xcframework"
 
 val cleanXcframework by tasks.creating(Exec::class) {
 
--- a/docs/Package.swift
+++ b/docs/Package.swift
@@ -1,0 +1,1 @@
+import PackageDescription


mikasdo (approved):

common/core/build.gradle.kts
```diff
@@ -88,7 +88,7 @@ android {
 
 apply(from = "../../android/scripts/maven-meta.gradle")
 
-val xcframeworkPath = "build/bin/xcframework/MaasCore.xcframework"
+val xcframeworkPath = "../../ios/MaasCore/Sources/MaasCore/Core.xcframework"
```
justasm, line 91: Out of curiosity, is there any technical benefit to moving the location of the xcframework?

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 // MARK: - Setup
 
-private let environment: Environment = .development
+private let environment: Environment = .production
```
justasm, line 8: I really want to avoid a persistent git diff when working on the iOS SDK locally.
1. You mentioned that environment values are not possible, but there is conflicting evidence on the internet 😄 [swift-evolution 0273](https://github.com/apple/swift-evolution/blob/main/proposals/0273-swiftpm-conditional-target-dependencies.md) mentions that "Swift 5.3 supports platform conditionals, but not configuration conditionals". [This Swift forum post](https://forums.swift.org/t/how-to-use-a-custom-build-configuration-in-a-swift-package/36259) implies that `.debug` and `.release` configurations are available (could we use these?). [This article](https://www.oreilly.com/library/view/extending-swift-values/9781492048817/ch04.html) suggests that environment values can be used - but maybe just not in `Package.swift`. [This gist](https://gist.github.com/Sorix/21e61347f478ae2e83ef4d8a92d933af) shows a concrete example of using environment values, but I have no idea if it actually works. @Ignelios @mikasdo where is the truth? 😄
1. What happens if this is `.development` by default? Is the issue that SDK users that include it via SPM get the `.development` configuration?

common/core/build.gradle.kts
Ignelios: When we're trying to fetch MaasCore for development we're using `.package(path: "/ios/MaasCore")` which requires as to create a target inside `Package.swift` for `MaasCore` and path for target must by inside package directory.

PR #99, opened 2021-01-25
Fix build of MaasCore, MaasComponents & MaasTheme

justasm:
Previously the project would not build when opening `MaasCore/Package.swift`.


mikasdo (approved):

Ignelios:
This should be covered by [#97](https://github.com/trafi/maas-components/pull/97)

PR #101, opened 2021-01-26
Kotlin Flow to Swift Combine interop

justasm:
All API methods now expose a Combine `Publisher` in addition to a completion handler variant. To use, call

```swift
let publisher = whicheverApi.whicheverCall(arguments).publisher
publisher.sink { completion in
    // do on completion
} receiveValue: { value in
    // use value
}
```

To expose such publishers, we (for now, manually) need to
1. Wrap each API call into a `CFlow` type in the `common:core` `iosMain` sourceset, e.g.

```kotlin
fun UsersApi.createOrGetUser(profile: Profile? = null): CFlow<ApiResult<User>> =
    cFlow { emit(createOrGetUser(profile)) }
```

2. Add a Swift extension to `CFlow` for each response type. See `CFlow+Publisher.swift`, e.g.

```swift
extension CFlow where T == ApiResult<User> {
    var publisher: AnyPublisher<User, ApiError> { cflowToPublisher(self) }
}
```

`CFlow` is a variant of Kotlin's `Flow` type (analogous to Rx Observable or Swift Publisher) which exposes a `watch` method and allows to listen for newly emitted elements using a callback.

Additionally,
- `ApiResult<T>` type `T` is no longer nullable in Swift


common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
```diff
@@ -0,0 +1,28 @@
+@file:Suppress("unused", "EXTENSION_SHADOWED_BY_MEMBER")
```
Ignelios, line 1: What does these suppress does exactly ? 🤗

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.trafi.core.model.AutoCompleteLocations
+import com.trafi.core.model.LatLng
+import com.trafi.core.model.Location
+import com.trafi.core.model.ReverseGeocodeResponse
+
+fun LocationsApi.search(
+    query: String,
+    coordinate: LatLng? = null,
+): CFlow<ApiResult<AutoCompleteLocations>> =
+    cFlow { emit(search(query, coordinate)) }
```
Ignelios, line 18: We'll generate these `fun` in near future right? 🤔

ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
```diff
@@ -0,0 +1,8 @@
+enum ApiError: Swift.Error, Equatable, Hashable {
```
Ignelios, line 1: Very nice! 👍

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
Ignelios, line 1: What does this exactly ? 🤗

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
Ignelios, line 18: We'll generate these fun in near future right ? 🤔

ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
Ignelios, line 1: Very nice! 👍

ios/MaasCore/Sources/MaasCore/Api/Utils/CFlow+Publisher.swift
```diff
@@ -0,0 +1,31 @@
+import Combine
```
Ignelios, line 1: Possible place to keep all generated flow extensions. 👍

Ignelios (approved):

common/core/src/commonMain/kotlin/com/trafi/core/ApiResult.kt
```diff
@@ -1,6 +1,13 @@
...
-    class Success<T>(val value: T) : ApiResult<T>()
-    class Failure<T>(val exception: Throwable) : ApiResult<T>()
+/**
+ * A discriminated union that encapsulates a successful result with a value of type [T]
+ * or a failure with an arbitrary [Throwable] exception.
+ *
+ * The generic upper bound [Any] is specified due to limitations in
+ * [interop with Objective-C](https://kotlinlang.org/docs/reference/native/objc_interop.html#generics).
+ */
+sealed class ApiResult<out T : Any> {
```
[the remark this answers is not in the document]
justasm: Yes, the generic type will no longer be nullable.

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
justasm: They suppress warnings from the Kotlin compiler. So they're not necessary, but without suppressing them the compiler & IDE will display warnings such as
- `unused` - "This function is not used"
- `EXTENSION_SHADOWED_BY_MEMBER` - "This extension function is shadowed by a member function of the same name"

The first is irrelevant - we know these functions are not used within the Kotlin code, but they are used from Swift in the published xcframework.
The second is also not a problem - in Kotlin e.g. `suspend fun UsersApi.createOrGetUser()` and `fun UsersApi.createOrGetUser(): CFlow` may clash, but when translated to Objective-C / Swift their signatures are different - `func createOrGetUser(completionHandler: ..)` and `func createOrGetUser(): CFlow`.

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
justasm: Yes, that's the idea

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
justasm: See [previous answer](https://github.com/trafi/maas-components/pull/101#discussion_r565947786)

common/core/src/iosMain/kotlin/com/trafi/locations/LocationsApi.kt
justasm: Only robots will have "fun" 😂

PR #102, opened 2021-01-27
Add new corebinary checksum and url.

Ignelios:
Deleted all current pre-releases now we'll only update a first one `(0.1.0-dev01)` with new `CoreBinary.xcframework` and checksum.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -99,7 +99,7 @@ private extension Environment {
             return [
                 .binaryTarget(
                     name: "CoreBinary",
-                    url: "https://github.com/trafi/maas-components/releases/download/0.1.0-dev01/CoreBinary.xcframework.zip",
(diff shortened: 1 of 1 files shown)


justasm:
Can we start with dev03? The dev02 version of the Android SDK is already released publicly and cannot be taken back 🙂

Ignelios:
Yeah sure we can keep this at dev03 👍

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 // MARK: - Setup
 
-private let environment: Environment = .development
+private let environment: Environment = .production
```
Ignelios, line 8: Changing back to production, because we cant use spm xcode because spm has local package dependencies. 😢

justasm (approved):

PR #103, opened 2021-01-28
Reuse common String to Color parsing logic

justasm:


ios/MaasCore/Sources/MaasTheme/Theme/Colors.swift
```diff
@@ -156,6 +156,25 @@ public struct GrayScalePalette {
...
+    }
+}
+
+public extension Color {
+    init(hex: String?) {
+        let uiColor: UIColor = ColorKt.toColor(hex, fallback: 0xFF_000000).color
+        self.init(uiColor)
+    }
+}
+
```
Ignelios, line 177: I have a felling that we should create a separate `*.swift` for these and keep all extensions in one place. What do you think? 🤔

justasm: Please do 🙏

justasm: Kept `Color.init(hex: String?)` here as we want to keep the `UiColor.color` extension internal.

mikasdo (approved):

PR #104, opened 2021-01-28
Introduce kotlin propertywrapper

Ignelios:
before
```swift
@Environment(\.currentTheme) var theme
var constants: Kotlin<SomeConstants> { Kotlin(SomeConstants(theme: theme)) }
```

after
```swift
@Themeable(SomeConstants.init) var constants
```


mikasdo:
@Ignelios what do you think about this kind of declaration?
```swift
@Themeable(SomeConstants.init) var constants
```

justasm (approved):
Oh man, I love it! 😍 Really elegant 👏

PR #105, opened 2021-01-29
Parse API response error body

justasm:
- return a distinct error case for 401 unauthorized errors (`ApiError.unauthorized` in iOS, `ApiResult.Failure.Unauthorized` in common and Android)
- parse and return the API's `Error` response body when an error response is received
- update `LocationsApi` and `RoutesApi` to use `ApiConfiguration`
- basic unit test setup for `common:core` Kotlin multiplatform tests

### DIFF
--- a/common/core/src/commonMain/kotlin/com/trafi/maas/transit/ui/internal/NearbyTransitScheduleBadgesConstants.kt
+++ b/common/core/src/commonMain/kotlin/com/trafi/maas/transit/ui/internal/NearbyTransitScheduleBadgesConstants.kt
@@ -1,3 +1,5 @@
+@file:Suppress("UNUSED_PARAMETER")
+
 package com.trafi.maas.transit.ui.internal
 
 import com.trafi.ui.theme.internal.CurrentTheme
--- a/ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
+++ b/ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
@@ -1,8 +1,20 @@
+/**
+ - Tag: ApiError
+ */
 enum ApiError: Swift.Error, Equatable, Hashable {
-    case httpError
-    case parseError
-    case error(String)
-    case unknownError
-    
-    case apiResultFailure
+    /**
+     The server returned 401.
+     - Parameter error: optional error message.
+     */
+    case unauthorized(error: Error?)
+    /**
+     The server returned an error.
+     - Parameter error: optional error message.
+     */
+    case error(error: Error?)
+    /**
+     A generic failure not covered by the more specific [ApiError](x-source-tag://ApiError) types.
+     - Parameter developerMessage: optional message to help identify the error.
+     */
+    case failure(developerMessage: String?)
 }
--- a/ios/MaasCore/Sources/MaasCore/Api/Utils/ApiKotlinFlowPublisher.swift
+++ b/ios/MaasCore/Sources/MaasCore/Api/Utils/ApiKotlinFlowPublisher.swift
@@ -27,14 +27,21 @@ private extension ApiKotlinFlowPublisher {
             self.subscriber = subscriber
             
             self.closable = flow.watch { result in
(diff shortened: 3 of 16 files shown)


ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
```diff
@@ -1,8 +1,20 @@
+/**
+ - Tag: ApiError
+ */
 enum ApiError: Swift.Error, Equatable, Hashable {
```
Ignelios, line 4: `ApiError` should be `public enum` because we won't be able to access it when we install SPM dependencies in to custom project.

ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
```diff
@@ -1,8 +1,20 @@
+/**
+ - Tag: ApiError
```
Ignelios, line 2: Very nice job! With descriptions. 🎖️

ios/MaasCore/Sources/MaasCore/Api/Utils/ApiKotlinFlowPublisher.swift
```diff
@@ -27,14 +27,21 @@ private extension ApiKotlinFlowPublisher {
...
                     let _ = subscriber.receive(success.value)
-                } else if let failure = result as? ApiResultFailure<T> {
-                    // TODO: failure.exception as ApiError ???
-                    subscriber.receive(completion: .failure(ApiError.apiResultFailure))
-                } else {
-                    let message = "ApiResult: unable to parse to ApiResultSuccess or ApiResultFailure, Expecting type: \(T.self)"
-                    subscriber.receive(completion: .failure(ApiError.error(message)))
+                case let failure as ApiResultFailureUnauthorized<T>:
+                    subscriber.receive(completion: .failure(ApiError.unauthorized(error: failure.error)))
+                case let failure as ApiResultFailureError<T>:
```
Ignelios, line 35: Can we use this like that ? 🤔 
```suggestion
                    subscriber.receive(completion: .failure(.unauthorized(error: failure.error)))
                case let failure as ApiResultFailureError<T>:
```

common/core/src/commonMain/kotlin/com/trafi/maas/transit/ui/internal/NearbyTransitScheduleBadgesConstants.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("UNUSED_PARAMETER")
```
Ignelios, line 1: What does this exactly ? 🤔

Ignelios (approved):
LGTM 👍 just a few things. Noticed that all `CFlow` extensions should be declared as `public` which leads us to chain of a `classes/extension` which should be declared as public. We need that to have a full access to `ApiError or/and .publisher` when we're importing Maas Package.

ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
justasm: OK, unrelated to the scope of this PR

ios/MaasCore/Sources/MaasCore/Api/Utils/ApiKotlinFlowPublisher.swift
justasm: We sure can! Swift 😍

common/core/src/commonMain/kotlin/com/trafi/maas/transit/ui/internal/NearbyTransitScheduleBadgesConstants.kt
justasm: Suppresses warnings from the Kotlin compiler that the class has members that are declared but not used.

justasm:
Good point regarding the visibility of extensions 👍 But it's not related to the changes in this PR, so let's address it separately.

PR #107, opened 2021-02-01
Api error swift

mikasdo:
Refactor to make more readable and easier to understang converting from some ApiResultFailure to ApiError


common/core/src/commonMain/kotlin/com/trafi/core/ApiResult.kt
```diff
@@ -45,7 +45,7 @@ sealed class ApiResult<out T : Any> {
                     val error: Error? = try {
                         Json.decodeFromString(throwable.response.content.readRemaining().readText())
                     } catch (e: SerializationException) {
-                        null
+                        Error(e.message ?: "unable to decode error, ktorFailure")
```
justasm, line 48: I'm not certain we want to construct our own `Error` here 🤔 `Error` is an error returned from our API. As far as I know not all API error responses return an `Error`. Is it an error for our API to return an error without an `Error`? 😄 If the API trips because of an internal server error and we don't get an `Error`, is it correct for our `Failure.Error` to contain a `throwable` and an unassociated `error` that is related to our parsing logic?

ios/MaasCore/Sources/MaasCore/Api/Utils/ApiKotlinFlowPublisher.swift
```diff
@@ -27,21 +27,10 @@ private extension ApiKotlinFlowPublisher {
...
-                case let failure as ApiResultFailureGeneric<T>:
-                    subscriber.receive(completion: .failure(.failure(developerMessage: failure.throwable.message)))
-                case let failure as ApiResultFailure<T>:
-                    let message = "ApiResult: could not parse \(failure) as sub-type of ApiResultFailure. Failure message: \(failure.throwable.message ?? "nil")"
-                    subscriber.receive(completion: .failure(.failure(developerMessage: message)))
-                default:
-                    let message = "ApiResult: unable to parse \(result) as ApiResultSuccess or ApiResultFailure"
-                    subscriber.receive(completion: .failure(.failure(developerMessage: message)))
+                } else {
+                    subscriber.receive(completion: .failure(ApiError(result)))
```
justasm, line 33: Nice 👍

justasm (approved):

common/core/src/commonMain/kotlin/com/trafi/core/ApiResult.kt
mikasdo: Initial idea was to provide more visibility when debugging in swift. Ignas was receiving Failure with error == nil, so just to be sure if nil was coming from this place in shared code, we put some message there, recompile xcframework, and then realise e that the assumption was correct.. so just to have a clear distinction what kind of error, instead of error == nil, I suggest to create our  own error here

justasm: As discussed in person - it is a fact that our API sometimes doesn't return an `Error`, so I'd argue the most straightforward way to represent this scenario is to represent it as nullable.

All `ApiResult.Failure` types that have an associated HTTP response also have a `failure.httpStatusCode`, and all `ApiResult.Failure` types also have the underlying exception as `failure.throwable`. It may make sense to use the `failure.throwable.message` when the `failure.error?.developerMessage` is nil, but this can be done at the higher level of abstraction, e.g. when mapping to the Swift `ApiError`.

Ignelios (approved):

PR #111, opened 2021-02-07
ErrorsError -> Error

mikasdo:
- Fix Error and ErrorsError confusion in generated Api models


mikasdo:
Do you agree with "PlatformError"? Or maybe "ApiError" / "NetworkError" / "???"

justasm (approved):
How is the model named in the Swagger schema? We discussed renaming it back to just `Error` iirc. We'll want to generate the models automatically, so it would be most straightforward for the name to match that in the schema.

mikasdo:
Yes, we discussed renaming to Error, but I thought as we are making changes we might resolve the conflict on Swift side.. The naming in schema can be adjusted if I'm correct, because the Error model was assigned schemaId "ErrorsError".
If I'm correct, we could adjust the naming in schema to our needs, and the generated models would follow 🤔

justasm:
We can adjust the naming in the schema, but currently it's part of the platform's public API, so it affects other clients using the platform-api directly. I see that with https://github.com/trafi/trafi-platform-api/pull/424/ it was renamed back to `Error` in the schema.

I see two possible approaches
1. add support for different schema ids for different Swaggers in the platform-api (so we can have our SDK-specific name but not affect other clients, may get confusing in the long run though)
1. add support in our (yet to be made) code generator to map certain schema names to others

Ignelios (approved):

mikasdo:
Renaming back to `Error`, will keep in mind code generator

PR #115, opened 2021-02-11
Back to development, importing SwiftUI.State in new components.

Ignelios:
- Setting environment back to `.development`
- `import struct SwiftUI.State` inside Button and Spinner views.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 
 // MARK: - Setup
 
-private let environment: Environment = .production
+private let environment: Environment = .development
 
 let package = Package(
     name: "Maas",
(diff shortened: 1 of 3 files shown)


mikasdo (approved):

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 // MARK: - Setup
 
-private let environment: Environment = .production
+private let environment: Environment = .development
```
justasm, line 8: What's the reasoning here? 🤔

justasm (approved):

Package.swift
Ignelios: With some other pull request by mistake we reversed this to production. So but we still are using only a development environment.

PR #117, opened 2021-02-11
Expose 403 Forbidden errors

justasm:
Expose 403 as Kotlin `ApiResult.Failure.Forbidden` and Swift `ApiError.forbidden`. This typically corresponds to an invalid API key or when the API key has not subscribed to the API.

### DIFF
--- a/ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
+++ b/ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
@@ -17,7 +17,13 @@ class ApiConfiguration: ApiConfig {
 
 public protocol ApiConfig {
     var baseUrl: String { get }
+    /**
+     - Tag: ApiConfig.apiKey
+     */
     var apiKey: String { get }
+    /**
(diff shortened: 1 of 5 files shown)


ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
```diff
@@ -17,7 +17,13 @@ class ApiConfiguration: ApiConfig {
 
 public protocol ApiConfig {
     var baseUrl: String { get }
+    /**
+     - Tag: ApiConfig.apiKey
```
mikasdo, line 21: What is the purpose of those tags? better generated documentation?

mikasdo (approved):

ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
justasm: So we can link from other source documentation using `x-source-tag://`. See e.g. the inline documentation for `ApiError.forbidden`.

PR #118, opened 2021-02-15
Update api config and add token refresh functionality

mikasdo:


ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
```diff
@@ -1,30 +1,16 @@
...
-        return ""
-    }
+    ///  - Tag: ApiConfig.getIdToken
+    func getIdToken() -> String?
     
-    var logger: Logger? { ApiLogger.default }
-}
+    /// - Tag: ApiConfig.getIdToken
+    ///  must ensure that when completion is called, [getIdToken()](x-source-tag://ApiConfig.getIdToken) will return refreshed token
+    func refreshIdToken(completion: @escaping (String?) -> ())
```
justasm, line 12: I think it's a sensible choice to have a global way to configure `getIdToken` and `refreshIdToken` 👍 In the future, if we want to support more unauthenticated use cases (transit, route search, etc), we may make these optional.

ios/MaasCore/Sources/MaasCore/Extensions/Publishers+Retry.swift
```diff
@@ -0,0 +1,105 @@
...
+            self.upstream
+                .catch { e -> AnyPublisher<Output, Failure> in
+                    guard failureShouldRetry(e) else {
+                        return Fail<Output, Failure>(error: e)
+                            .eraseToAnyPublisher()
+                    }
+                    
+                    return recoveryPublisher
+                        .catch { _ in Fail(error: e) }
+                        .flatMap { _ in upstream }
```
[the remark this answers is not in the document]
mikasdo: yes, if you need to retry multiple times, you can wrap the resulting publisher, with `.retry(2)` ;)

ios/MaasCore/Sources/MaasCore/Api/Utils/RefreshToken+Publishers.swift
```diff
@@ -0,0 +1,31 @@
...
+
+private var isRefreshing = false
+private var activeRefreshTokenPublisher: AnyPublisher<Void, Never>!
+
+extension ApiConfig {
+    
+    func refreshTokenPublisher() -> AnyPublisher<Void, Never> {
+        Deferred { [unowned self] () -> AnyPublisher<Void, Never> in
+            if !isRefreshing {
+                isRefreshing = true
```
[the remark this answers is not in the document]
mikasdo: used NSLock. Deferred has very little documentation, I can't be sure that combine's `.subscribe(on` will affect the execution of code inside deferred, so for now sticking with what seems to be working well in our whitelabel apps :)

ios/MaasCore/Sources/MaasCore/Api/Utils/RefreshToken+Publishers.swift
```diff
@@ -0,0 +1,35 @@
...
+        
+    func refreshTokenPublisher() -> AnyPublisher<Void, Never> {
+        Deferred { [unowned self] () -> AnyPublisher<Void, Never> in
+            
+            access.lock(); defer { access.unlock() }
+            
+            if !isRefreshing {
+                isRefreshing = true
+                activeRefreshTokenPublisher = self.performTokenRefresh()
+                    .handleEvents(receiveCompletion: { _ in isRefreshing = false })
```
justasm, line 17: Need to think through whether this assignment should also be guarded, e.g. in a concurrent scenario, what happens when
1. (thread 1) enters Deferred block, if !isRefreshing returns false, i.e. `isRefreshing == true`
2. (thread 2) `activeRefreshTokenPublisher` completes and fires `receiveCompletion`, setting isRefreshing to false
3. (thread 1) returns (the already completed) `activeRefreshTokenPublisher`

Will a new subscriber to `activeRefreshTokenPublisher` after it has already completed get an emission?

mikasdo: `activeRefreshTokenPublisher ` is constructed from a `Future`, all subsequent subscriptions to a publisher, that is constructed from a future will receive the same value, and finish events.
example shows 2️⃣ subscription that subscribes after `Future` is completed, receives the same value immediately, without any delay.
 
https://user-images.githubusercontent.com/70565612/108203286-f2bd7600-712a-11eb-9090-ed2974c3d18b.mp4

mikasdo: Just to keep it more clear - moved the `isRefreshing = false` to token refresh completion handling with the same NSLog protection

ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
```diff
@@ -1,30 +1,16 @@
...
-        return ""
-    }
+    ///  - Tag: ApiConfig.getIdToken
+    func getIdToken() -> String?
     
-    var logger: Logger? { ApiLogger.default }
-}
+    /// - Tag: ApiConfig.refreshIdToken
+    ///  must ensure that when completion is called, [getIdToken()](x-source-tag://ApiConfig.getIdToken) will return refreshed token
+    func refreshIdToken(completion: @escaping (String?) -> ())
```
justasm, line 12: One final thought - what do you think about making this method optional? i.e. if the SDK user specifies it, they *opt-in* to automatic retry on unauthorized, otherwise they *opt-out*. I would like the retry functionality we provide to be explicit and optional, if possible, to minimize potentially surprising behavior.

justasm (approved):
Nice work 👍

ios/MaasCore/Sources/MaasCore/Api/ApiConfiguration.swift
mikasdo: Good idea 👍  added a default implementation, to make declaring it optional

PR #110, opened 2021-02-05
Button with icon / spinner

mikasdo:
https://user-images.githubusercontent.com/70565612/107225662-bc447480-6a21-11eb-8b10-8d44ef7a3424.mp4

TODO: 
 - convenience api for setting titles for different states ??
 
 *Note*: (as discussed in person) Android implementation will be in a separate PR

PR #120, opened 2021-02-17
Update Compose Button implementation

justasm:
Tweak `disabledBackgroundColor`.

Add support for
- icon
- small format
- wrapped format
- loading state

Keeps the Compose / Android implementation in line with SwiftUI #110
Didn't manage to nail down all the animations, there's still some jankiness, but decided it was good enough for now.

https://user-images.githubusercontent.com/1081815/108312933-141c7180-71c0-11eb-9339-bf728b51adb0.mp4

### DIFF
--- a/android/ui-android/src/main/java/com/trafi/ui/component/Button.kt
+++ b/android/ui-android/src/main/java/com/trafi/ui/component/Button.kt
@@ -1,15 +1,37 @@
 package com.trafi.ui.component
 
 import androidx.annotation.VisibleForTesting
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.defaultMinSizeConstraints
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.CircularProgressIndicator
+import androidx.compose.material.Icon
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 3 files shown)


common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/ButtonConstants.kt
```diff
@@ -11,12 +11,11 @@ class ButtonConstants(theme: CurrentTheme) {
 
     val defaultBackgroundColor = theme.colorPalette.primary
     val defaultContentColor = theme.colorPalette.onPrimary
-    val disabledBackgroundColor = theme.colorPalette.primary.alpha(0.12f)
+    val disabledBackgroundColor = theme.colorPalette.primary.alpha(0.30f)
```
justasm, line 14: This seems to match the latest design in Figma.

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/ButtonConstants.kt
```diff
@@ -26,6 +25,8 @@ class ButtonConstants(theme: CurrentTheme) {
 
     val spaceBetween = SpacingScale.xs
     val paddingHorizontal = SpacingScale.md
+    val paddingVertical = 12.dp
+    val paddingVerticalSmall = 6.dp
```
justasm, line 29: cc @mikasdo I wasn't sure if it was appropriate to add a `.padding(.vertical, ..)` to `Button.swift`.

android/ui-android/src/main/java/com/trafi/ui/component/Button.kt
```diff
@@ -20,40 +42,138 @@ import com.trafi.ui.theme.lightColors
...
         )
     }
 }
 
+@Composable
+private fun ButtonConstants.contentColor(enabled: Boolean): Color =
+    if (enabled) defaultContentColor else disabledContentColor
+
+@Composable
+public fun SecondaryButton(
```
justasm, line 123: It seems to be in line with other Compose components (e.g. Material components) to declare common variants as separate components that delegate to the more generic component (as opposed to e.g. using an enum). I've added `SecondaryButton` and `TertiaryButton` with that in mind.

cc @mikasdo @Ignelios

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/ButtonConstants.kt
mikasdo: I think it's either height, or padding 🤔 but when we have scaling fonts, we can use height as min height and add padding to scaled text when it gets bigger

android/ui-android/src/main/java/com/trafi/ui/component/Button.kt
Ignelios: Nice i think we should implement this too.

common/core/src/commonMain/kotlin/com/trafi/ui/component/internal/ButtonConstants.kt
justasm: Yeah, the Compose implementation interprets the height as "min height".

Ignelios (approved):

PR #125, opened 2021-02-22
Swift button variants: primary / secondary / tertiary

mikasdo:
Screenshot 2021-02-22 at 09 21 26


ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
```diff
@@ -0,0 +1,208 @@
...
+    let disabledColor: Color
+    
+    func stateful(isActive: Bool, isEnabled: Bool) -> Color {
+        if !isEnabled { return disabledColor }
+        if isActive { return activeColor }
+        return color
+    }
+    
+    static func just(_ color: Color) -> StatefulColor {
+        StatefulColor(color: color, activeColor: color, disabledColor: color)
```
justasm, line 39: Nice 👍

ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
```diff
@@ -0,0 +1,208 @@
...
+    let input: Button.InputType
+    
+    @Environment(\.isEnabled) var isEnabled
+    
+    var body: some View {
+        Button(
+            input.text,
+            icon: input.icon,
+            foregroundColor: colors.foreground.stateful(isActive: input.isLoading, isEnabled: isEnabled),
+            backgroundColor: colors.background.stateful(isActive: input.isLoading, isEnabled: isEnabled),
```
justasm, line 13: So `Button.InputType` colors are completely ignored 🤔 I think they should take precedence and we only fall back to these colors as defaults.

ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
```diff
@@ -0,0 +1,208 @@
...
+        if isActive { return activeColor }
+        return color
+    }
+    
+    static func just(_ color: Color) -> StatefulColor {
+        StatefulColor(color: color, activeColor: color, disabledColor: color)
+    }
+}
+
+public struct PrimaryButton: View {
```
justasm, line 43: Can the standard `Button` be "primary", i.e. have the primary button defaults?

justasm (approved):
👏

ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
justasm: Keep in mind that if the SDK user specifies colors in `InputType`, they did so explicitly at the call-site, so it would be unexpected for them to not be respected.

mikasdo: I left no colors in `init` for ButtonVariants

ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
mikasdo: Standard button only has content and background colors, PrimaryButton has colors for disabled / active states as well

Ignelios:
So these button types don't support swappable ? 🤔 Only Button ?

mikasdo:
Correct. You swap `Button` and all button variants will use your swapped implementation

Ignelios (approved):

ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
justasm: Ah, OK, I see, so we offer no way to customize colors for `ButtonVariants`. Not obvious to me that we won't want to support custom colors for `PrimaryButton`, but seems sensible for now 👍

ios/MaasCore/Sources/MaasComponents/Components/ButtonVariants.swift
justasm: I see 👍

PR #128, opened 2021-02-23
No more exported

Ignelios:


ios/MaasCore/Sources/MaasCore/Api/ApiLogger.swift
```diff
@@ -1,20 +1,21 @@
...
-
-extension OSLog {
-    
-    private static var subsystem = Bundle.main.bundleIdentifier!
-    
-    static let MaasComponentsApi = OSLog(subsystem: subsystem, category: "MaasComponents")
-}
+import CoreBinary
+//
+//class ApiLogger : CoreBinary.Logger {
```
mikasdo, line 4: This could be deleted as it is an example, ant there is one in Login sample

Package.swift
```diff
@@ -104,10 +110,10 @@ private extension Environment {
                 .binaryTarget(
                     name: "CoreBinary",
                     url: "https://github.com/trafi/maas-components/releases/download/0.1.0-dev03/CoreBinary.xcframework.zip",
-                    checksum: "08d2bc473fe1ea73d9d389b8b491ecd378fde82c49d24bc1579804f0e8ed978b"),
+                    checksum: "419f3c06d6dc73f042794b81eb7c92e922077f11a0253aa04817fa20938dc118"),
```
mikasdo, line 113: if you have uploaded a new binary, shouldn't we have used a new tag as well? dev04?

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 // MARK: - Setup
 
-private let environment: Environment = .development
+private let environment: Environment = .production
```
mikasdo, line 8: .production works good for development? or this is for easier integration in whitelabel?

mikasdo (approved):

Package.swift
Ignelios: We're uploading xcframework in to dev-03 instead of creating new one.

justasm: We discussed that we will reuse the same tag until someone actually uses it. But it will be necessary to move the dev03 tag to this commit.

Package.swift
Ignelios: Its for integration in to WL because we're about to use this like production 🙌

ios/MaasCore/Sources/MaasCore/Api/ApiLogger.swift
Ignelios: Sure i'll delete this. 🙇

Package.swift
Ignelios: So did i get it right we increasing tag? 🤔

mikasdo: tag stays the same, but we will need to move tag to the commit after merge

PR #129, opened 2021-02-23
Platform error codes

mikasdo:
Introduce UsersError and MspError (errorCode(msp / users) in swift) for convenience when handling custom error codes


ios/MaasCore/Sources/MaasCore/Api/Utils/RefreshToken+Publishers.swift
```diff
@@ -41,8 +41,13 @@ extension Publisher where Failure == ApiError {
         retryWithRecovery(
             recoveryPublisher: Maas.apiConfig.refreshTokenPublisher(),
             failureShouldRetry: {
-                if case .unauthorized = $0 { return true }
-                else { return false }
+                switch $0 {
+                case .unauthorized: return true
+                case .errorCode(let code, _):
+                    return code == .user(.tokeninvalidorexpired) ||
+                        code == .user(.refreshtokeninvalidorexpired)
```
mikasdo, line 48: if we handle specific error codes tokeninvalidorexpired || refreshtokeninvalidorexpired should we still trigger token refresh on `401` - (.unauthorized)?

justasm: Seems like yes, based on yesterday's test where an error code is not returned in case the token is empty 🙂

justasm: I don't think we can handle `refreshtokeninvalidorexpired` in the SDK as we don't own the refresh token.

mikasdo: yesterday's test was flawed 😅 there i used a fake token corrupt, so the token wasn't actually expired.. today my token got old, and i received `tokeninvalidorexpired`

common/core/src/commonMain/kotlin/com/trafi/core/ErrorCode.kt
```diff
@@ -0,0 +1,42 @@
+package com.trafi.core
+
+sealed class ErrorCode {
+    enum class Msp(val value: String) {
```
justasm, line 4: Unless you extend the sealed class, they won't be part of the sealed class hierarchy 🙂
```suggestion
    enum class Msp(val value: String): ErrorCode() {
```

If we do this, then perhaps we can reuse `ApiResult.Failure.Error`? And add an `val errorCode: ErrorCode` property.

ios/MaasCore/Sources/MaasCore/Api/Utils/RefreshToken+Publishers.swift
justasm: But if SDK users provide an invalid (or empty, or nil) id token out of the box, should we still ask for a new token and retry?

ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
```diff
@@ -19,17 +19,26 @@ public enum ApiError: Swift.Error, Equatable, Hashable {
      - Parameter error: optional error message.
      */
     case error(error: CoreBinary.Error?)
+    
+    case errorCode(code: Code, error: CoreBinary.Error?)
+
```
Ignelios, line 24: Let's discard `case error(error: CoreBinary.Error?)` and use `case error(code: Code, error: CoreBinary.Error?)` instead.

mikasdo: but code will not always be there 🤔 I would prefer separate enum case instead of optional `Code` but there is room for improvement 🤔

ios/MaasCore/Sources/MaasCore/Api/Utils/RefreshToken+Publishers.swift
mikasdo: probably not, I think we should retry only when we get explicit error code, that token is expired 🤔  do you agree?

ios/MaasCore/Sources/MaasCore/Api/ApiError.swift
Ignelios: Lets make it optional then 😄

ios/MaasCore/Sources/MaasCore/Api/Utils/RefreshToken+Publishers.swift
justasm: OK 🙂

common/core/src/commonMain/kotlin/com/trafi/core/ErrorCode.kt
mikasdo: changed ErrorCode to `object`.
as discussed privately, leaving `MspError` and `UserError` for now

common/core/src/commonMain/kotlin/com/trafi/core/ApiResult.kt
```diff
@@ -52,6 +52,26 @@ sealed class ApiResult<out T : Any> {
...
         ) : Failure<T>(throwable)
 
+        /**
+         * The server returned an error covered by [ErrorCode.Users] types.
+         */
+        class UserError<T : Any>(
+            throwable: Throwable,
+            val httpStatusCode: Int,
+            val code: ErrorCode.Users,
+            val error: com.trafi.core.model.Error? = null,
```
justasm, line 62: `Error` is not nullable in these cases, right?

justasm (approved):
Will need to update the auth-firebase Android sample app as it doesn't compile without handling the new error types.

common/core/src/commonMain/kotlin/com/trafi/core/ApiResult.kt
mikasdo: absolutely!

justasm (approved):

